PR #1, opened 2018-11-09
Feature/add map component

Levia:
Add basic map component and style.


app/assets/stylesheets/components-vue/_map.scss
```diff
@@ -0,0 +1,161 @@
...
+
+//--------------------------------------------------
+// infographic map
+//--------------------------------------------------
+  &--infographic {
+    height: rem-calc(180);
+
+    @include breakpoint($small) { height: rem-calc(300); }
+    @include breakpoint($medium) { height: rem-calc(480); }
+  }
```
stacytalbot: Infographic styles not needed for this proj - delete

app/assets/stylesheets/utilities/_variables.scss
```diff
@@ -15,8 +15,11 @@ $white: #ffffff;
 $grey-light: #f4f4f4;
 $grey: #7f7f7f;
 $grey-dark: #444444;
+$grey-xdark: #202020;
 $black: #000000;
 
+$green: #71A32B;
```
stacytalbot: Not for this proj - delete

app/assets/stylesheets/components-vue/_map.scss
```diff
@@ -0,0 +1,161 @@
...
+            border: solid 1px $white;
+            border-radius: 100%;
+            margin-right: rem-calc(10);
+            width: rem-calc(22); height: rem-calc(22);
+
+            display: block;
+            position: relative;
+
+            &::after {
+              background-color: $green;
```
stacytalbot: ::after not for this proj - delete

app/assets/stylesheets/components-vue/_map.scss
```diff
@@ -0,0 +1,161 @@
...
+            background-color: $grey;
+            border: solid 2px $white;
+            border-radius: 100%;
+            margin-right: rem-calc(14);
+            width: rem-calc(30); height: rem-calc(30);
+
+            display: block;
+            position: relative;
+
+            &::after {
```
stacytalbot: not for this proj - delete

app/controllers/site_controller.rb
```diff
@@ -1,264 +1,91 @@
...
+  private
 
+  def load_global
+    @global = YAML.load(File.open("#{Rails.root}/lib/data/content/global.yml", 'r'))
+    @aichi_targets = YAML.load(File.open("#{Rails.root}/lib/data/content/aichi-targets.yml", 'r'))
+    @sdgs = YAML.load(File.open("#{Rails.root}/lib/data/content/sdgs.yml", 'r'))
+  end
+
+  def load_charts_data
     @chart_greatest_coverage = [
```
stacytalbot: This chart data needs to be different for every page

app/views/global/_map.html.erb
```diff
@@ -0,0 +1,7 @@
+<mapbox
+  global-coverage-title="<%= @global_coverage_title %>"
```
stacytalbot: can you name these the other way around?
- title-coverage
- title-protected

app/javascript/components/map/Mapbox.vue
```diff
@@ -0,0 +1,54 @@
...
+  </div>
+</template>
+
+<script>
+  export default {
+    name: 'mapbox',
+
+    props: {
+      globalCoverageTitle: String,
+      protectedTitle: String,
```
stacytalbot: Can you name these the other way around?

- titleGlobal
- titleProtected

app/controllers/site_controller.rb
```diff
@@ -1,264 +1,91 @@
...
-        percent: '50',
-      },
-      {
-        label: 'Russia',
-        value: '10',
-        percent: '10',
-      }
-    ]
-
-    @chart_protected_areas = [
```
stacytalbot: Protected area data is needed can you add back in?

app/controllers/site_controller.rb
Levia: Isn't that still dummy data?

stacytalbot: Yea, but it will need to be different for every page

app/controllers/site_controller.rb
Levia: You can find this in the shared method called with `before_action`: `load_charts_data`.
Line 3 and 88. PA data at line 117

app/controllers/site_controller.rb
Levia: :( Ok

Levia: Does it need to be different even now? I'm planning to make this different in the future when we have got the real data in of course

stacytalbot: No thats fine, as long as you are aware 👍

Levia: phew

PR #2, opened 2018-11-12
Feature/compute data

Levia:
## Description

* Define data structure for habitats
* Refactor and improve code in controller
* Add modules to deal with CartoDB
* Calculate global coverage for each habitat

## Known Issues

FE needs to be fixed to show the total area properly (numbers are too big)


app/controllers/site_controller.rb
```diff
@@ -79,6 +45,11 @@ def coldwater
 
   private
 
+  def load_habitat;
```
lucacug: I think this is a typo

lib/modules/postgis_utils.rb
```diff
@@ -0,0 +1,19 @@
+module PostgisUtils
+  SRID = 4326.freeze
+
+  def self.intersection(geom, field='the_geom', make_valid=false)
+    field = make_valid ? "ST_MakeValid(#{field})" : field
+   "ST_Intersects(#{geom_from_text(geom)}, #{field})"
```
lucacug: Just an incorrect indentation here

app/controllers/site_controller.rb
```diff
@@ -79,6 +45,11 @@ def coldwater
...
+    @habitat = Habitat.where(name: action_name).first
```
lucacug: the `action_name` variable doesn't seem been defined anywhere

app/views/global/_map.html.erb
```diff
@@ -1,8 +1,8 @@
 <mapbox
-  habitat="<%= @title %>"
-  title-global="<%= @global_coverage_title %>"
-  title-protected="<%= @protected_title %>"
-  :percentage-global=<%= @global_coverage %>
-  :percentage-protected=<%= @protected_percentage %>
+  habitat="<%= @habitat.name %>"
```
stacytalbot: This needs to be the title not the name

Levia: done

app/controllers/site_controller.rb
Levia: It's a rails helper that returns the name of the current action

app/controllers/site_controller.rb
Levia: Fixed, thanks!

lib/modules/postgis_utils.rb
Levia: Fixed this as well. Well spotted!

PR #3, opened 2018-11-13
Feature/map

stacytalbot:
Add map layers


Levia:
One thing I noticed is that you can't zoom out once you zoom in (double clicking).
Can we add the zoom controls somewhere on the map? If the user is not supposed to zoom we should get rid of the double clicking functionality as well I think

PR #5, opened 2018-11-14
Feature/populate the charts

drewpotter:
This is the initial PR for populating the first chart on the page which is the “Top five countries and territories with the greatest coverage…” of each various type of habitat.

I have added various methods to the Carto class and I have been careful to make sure that the cold corals have the actual count instead of the area considered.

I think the frontend might need modifying for the cold corals to reflect that the value is actually a count for this habitat rather than an area.


app/controllers/site_controller.rb
```diff
@@ -56,34 +56,51 @@ def load_global
     @sdgs = YAML.load(File.open("#{Rails.root}/lib/data/content/sdgs.yml", 'r'))
   end
 
+  def sum_country_areas(total_area_by_country)
+    country_total_area = {}
+    total_area_by_country.flatten.each do |country_data|
+      next if country_data["iso3"].include? "/" #remove areas which have multiple iso
+      next if country_data["iso3"].include? "ABNJ" #remove ABNJ
+      country_total_area[country_data["iso3"]] = country_total_area[country_data["iso3"]].nil? ?  country_data["sum"] : country_total_area[country_data["iso3"]] + country_data["sum"]
```
Levia: Perhaps you could simply this by first initializing the hash when empty and then just adding the sum, like:

```
country_total_area[country_data["iso3"]] ||= 0
country_total_area[country_data["iso3"]] += country_data["sum"]
```

Worth checking the correctness but it should be ok, and also easier to read I believe ;)

app/controllers/site_controller.rb
```diff
@@ -56,34 +56,51 @@ def load_global
...
+    if @habitat.name == "coldcorals"
+      total_value_by_country = c.total_points_by_country
+      total_value_by_country = sort_country_count(total_value_by_country)
+    else
+      total_value_by_country = c.total_area_by_country
+      total_value_by_country = sum_country_areas(total_value_by_country)
+    end
+
+    top_five_countries = total_value_by_country.sort_by {|_key, value| value}.last(5)
+    arbitrary_value = top_five_countries.last.last.to_f * 1.05
```
Levia: Could you please explain this multiplication? Not sure why multiplying by 1.05 🤔

lib/modules/carto.rb
```diff
@@ -30,6 +44,16 @@ def total_area_query
     query
   end
 
+  def total_area_by_country_query
+    query = "SELECT SUM(gis_area_k),iso3 FROM #{from} GROUP BY iso3 ORDER by SUM(gis_area_k) desc"
```
Levia: Creating a variable here is not necessary, as the method will always return the output of the last line of code inside it. So the string only is sufficient.

lib/modules/carto.rb
```diff
@@ -30,6 +44,16 @@ def total_area_query
...
+    query
+  end
+
+  def total_points_by_country_query
+    query = "SELECT COUNT(*),iso3 FROM #{from} GROUP BY iso3 ORDER by COUNT(*) desc"
```
Levia: Same here

Levia:
As an overall comment, could you please try to move the logic about fetching and formatting the data outside the controller? The habitat model should a good candidate for now

app/controllers/site_controller.rb
drewpotter: I spoke with Roger and we discussed that the arbitrary value would be 5% greater than the highest value. Unless I have misunderstood this?

app/controllers/site_controller.rb
drewpotter: Many thanks I will try this :)

app/controllers/site_controller.rb
stacytalbot: @drewpotter this is correct

Levia: Great, thanks both for claryfing!

lib/modules/carto.rb
drewpotter: Thanks!

lib/modules/carto.rb
drewpotter: Thanks!

drewpotter:
Thanks @Levia I think I have fixed all the remaining issues on this.

PR #8, opened 2018-11-15
Add type flag to the habitat

drewpotter:
This is the initial work to add a `type` flag to the habitat for the frontend.


app/controllers/site_controller.rb
```diff
@@ -48,6 +48,7 @@ def coldcorals
   def load_habitat
     @habitat = Habitat.where(name: action_name).first
     @habitat ||= Habitat.where(name: 'coralreef').first
+    @habitat_hash = @habitat.habitat_to_hash
```
stacytalbot: Would this not work if you did `@habitat = @habitat.habitat_to_hash`

drewpotter: Later on `load_charts_data` needs that `@habitat` variable though. Unless I rework that code too.

drewpotter: What I really hoped was to access the type from the actual `@habitat` model.

app/views/global/_greatest-coverage.html.erb
```diff
@@ -10,7 +16,7 @@
               <div class="chart__bar">
                 <p class="chart__bar-label no-margin">
                   <%= number_with_delimiter(item[:value]) %>
-                  <% unless item[:type] == 'points' then %>km<sup>2</sup><% end %>
+                  <% unless @habitat_type == 'points' then %>km<sup>2</sup><% end %>
```
lucacug: I think the keyword `then` is redundant and can be safely removed

PR #10, opened 2018-11-16
Import static csv data

drewpotter:
This is the initial work to import the static CSV data including a `StaticStat` model for storing the data. Please populate your database once it has been migrated with the following command:
```
rake import:staticstats
```


lib/tasks/import_staticstats.rake
```diff
@@ -0,0 +1,93 @@
...
+      end
+    end
+      
+    csv.close
+    
+  end
+  
+  def insert_static_stat(csv_file, kind, habitat, iso3, value)
+    return if iso3.include? "/"
+    return if iso3.include? "ABNJ"
```
lucacug: I think here you could do so something like
`return if ['ABNJ', '/'].include?(iso3)

lib/tasks/import_staticstats.rake
```diff
@@ -0,0 +1,93 @@
+require 'csv'
+
+namespace :import do
+  desc "import CSV data into database"
+  task :staticstats, [:csv_file] => [:environment] do |t, args|
```
lucacug: seems you are not using `t` or `args` anywhere in the code. I think you can safely omit them

lib/tasks/import_staticstats.rake
```diff
@@ -0,0 +1,93 @@
...
+    csv_headers = File.readlines(csv).first.split(",")
+    
+    staticstats_hash = {
+      iso3: csv_headers[0],
+      value: csv_headers[1].chomp
+    }
+    
+    CSV.parse(csv, headers: true, encoding: "utf-8") do |row|
+      csv_staticstats_row = row.to_hash
+      staticstats_row_hash = {}
```
lucacug: Also this hash seems not to being used anywhere

lib/tasks/import_staticstats.rake
```diff
@@ -0,0 +1,93 @@
...
+      elsif csv_file.include? "Total"
+        insert_static_stat(csv_file, "total", habitat, iso3, value)
+      end
+    end
+      
+    csv.close
+    
+  end
+  
+  def insert_static_stat(csv_file, kind, habitat, iso3, value)
```
lucacug: the csv_file argument of this method seems unuseful given that you are calling it hardcoding the `kind` of the file (line 50 and 52)

lib/tasks/import_staticstats.rake
```diff
@@ -0,0 +1,93 @@
...
+    return if iso3.include? "/"
+    return if iso3.include? "ABNJ"
+    Rails.logger.info "insert #{kind} value into habitat #{habitat}: #{value} into iso3: #{iso3}"
+    country = Country.find_by(iso3: iso3) rescue nil
+    habitat = Habitat.find_by(name: habitat) rescue nil
```
lucacug: I'm not sure you need `rescue nil` here given that you are checking that in the following lines...

lib/tasks/import_staticstats.rake
```diff
@@ -0,0 +1,93 @@
...
+
+    if country.nil? || habitat.nil?
+      Rails.logger.info "Cannot import #{kind} value into habitat #{habitat}: #{value} into iso3: #{iso3}"
+      return
+    end
+
+    static_stat = StaticStat.find_or_create_by(country_id: country.id, habitat_id: habitat.id)
+
+    if kind == "protected"
```
lucacug: I would have used a `case` statement but this is just my opinion, the code works well anyway

lib/tasks/import_staticstats.rake
drewpotter: Oh nice! Thanks @lucacug

lib/tasks/import_staticstats.rake
drewpotter: Good point! I guess it would be useful for passing arguments but we aren't in this case!

lib/tasks/import_staticstats.rake
drewpotter: Good point! Thanks!

lib/tasks/import_staticstats.rake
drewpotter: Yes exactly :)

lib/tasks/import_staticstats.rake
drewpotter: Cool! I will remove that

lib/tasks/import_staticstats.rake
drewpotter: Good point!

lib/tasks/import_staticstats.rake
drewpotter: I tried deleting those but I can't seem to get the rake task to run now!

lib/tasks/import_staticstats.rake
lucacug: Sorry, Andrew in this case I gave you a bad hint.
in this case, the `include?` for the array doesn't work because it checks the exact match, so you have to do something like:
`return if iso3.include? '/' || iso3.include? 'ABNJ' `
Sorry but I just discovered that the method `include?` has two different behaviours if  it deals with an array or string

drewpotter: No worries I will take a look!

PR #11, opened 2018-11-16
Initial work on the final stats

drewpotter:
This is the initial PR for the final stats which needed to be implemented. Please run the rake command to populate these final stats:

```
rake import:staticstats
```

### DIFF
--- a/app/controllers/site_controller.rb
+++ b/app/controllers/site_controller.rb
@@ -81,7 +81,13 @@ class SiteController < ApplicationController
       }
     end
 
-    top_five_protected_areas = StaticStat.where(habitat: @habitat).order("protected_percentage DESC").pluck(:country_id, :protected_percentage).to_a.first(5)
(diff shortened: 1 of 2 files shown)


app/controllers/site_controller.rb
```diff
@@ -81,7 +81,13 @@ def load_charts_data
       }
     end
 
-    top_five_protected_areas = StaticStat.where(habitat: @habitat).order("protected_percentage DESC").pluck(:country_id, :protected_percentage).to_a.first(5)
+    top_five_country_ids = top_five_countries.map do |country|
+      iso3 = country.first
+      country_id = Country.find_by(iso3: country.first).id
```
lucacug: You could remove the unused `iso3` variable or use that in the next line

drewpotter: Many thanks!

PR #14, opened 2018-11-19
Filter out terrestrial PAs from map

Levia:
## Description

Only show marine and coastal PAs on the map

### DIFF
--- a/app/javascript/components/map/Mapbox.vue
+++ b/app/javascript/components/map/Mapbox.vue
@@ -2,7 +2,7 @@
   <div class="map--statistics sm-counter-trigger">
     <div class="map__panel gutters">
       <p class="map__panel-title">{{ titleGlobal }}</p>
-      
+
       <p class="map__panel-stat no-margin">
         <counter sm-trigger="sm-counter-trigger" sm-target="sm-counter-target" :number="percentageGlobal"></counter>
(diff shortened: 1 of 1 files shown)


app/javascript/components/map/Mapbox.vue
```diff
@@ -134,7 +134,12 @@
         let sqlArray = []
 
         tables.forEach(table => {
-          sqlArray.push(`SELECT cartodb_id, the_geom, the_geom_webmercator FROM ${table}`)
+          if(table == process.env.WDPA_POLY_TABLE || table == process.env.WDPA_POINT_TABLE) {
+            sqlArray.push(`SELECT cartodb_id, the_geom, the_geom_webmercator FROM ${table} WHERE marine::INT > 0`)
+          }
+          else {
+            sqlArray.push(`SELECT cartodb_id, the_geom, the_geom_webmercator FROM ${table}`)
+          }
```
stacytalbot: Would prob be neater to write as follows

```
const sqlQuery = `SELECT cartodb_id, the_geom, the_geom_webmercator FROM ${table}`

      if(table == process.env.WDPA_POLY_TABLE || table == process.env.WDPA_POINT_TABLE) {
        sqlQuery += ' WHERE marine::INT > 0'
      }
      
      sqlArray.push(sqlQuery)
```

PR #24, opened 2018-12-18
Feature/refactor to single page

stacytalbot:
- Update Citation text in the footer to match Charlottes new design in Trello
- Bold the first paragraph in the header
- Refactor the whole site down from 5 pages to 1
- Remove the 'page name' from the site title because it is now only one page and this can't be updated via js
- Delete all files no longer used
- Leave in code for cartoDB in case it is reinstated at a later date


app/controllers/site_controller.rb
```diff
@@ -1,62 +1,82 @@
...
-
-  def coldcorals
-    @data = YAML.load(File.open("#{Rails.root}/lib/data/content/coldwater.yml", 'r'))
-
-    @commitments = [
-      @aichi_targets,
-      @sdgs,
-      @data['other_targets']
-    ]
+    @nav = [
```
Levia: Since this is static data, I think we could put this into a constant or a YML file.

app/controllers/site_controller.rb
```diff
@@ -1,62 +1,82 @@
...
+        theme: 'theme--pink',
+        title: 'Cold-water corals'
+      }
+    ].to_json
+
+    content = YAML.load(File.open("#{Rails.root}/lib/data/content/#{@habitat.name}.yml", 'r'))
+    aichi_targets = generateImageUrls @aichi_targets
+    sdgs = generateImageUrls @sdgs
+
+    @habitatData = {
```
Levia: We could extract this into a serialiser to improve the readability of the controller.
If not the whole hash, at least the `map` key seems the most suitable part for this.

app/controllers/site_controller.rb
```diff
@@ -67,6 +87,14 @@ def load_global
     @sdgs = YAML.load(File.open("#{Rails.root}/lib/data/content/sdgs.yml", 'r'))
   end
 
+  def generateImageUrls targets
+    targets['list'].each do |target|
```
Levia: You should be able to use `map` instead of `each` so you won't need to return `targets` at the end. `map` returns the modified array.

stacytalbot: If i don't return targets, it doesn't work

Levia: Ah, I see the point now. That's fine then!

app/controllers/site_controller.rb
stacytalbot: 👍

app/controllers/site_controller.rb
stacytalbot: @Levia my first serializer - be gentle

app/serializers/habitats_serializer.rb
```diff
@@ -0,0 +1,59 @@
...
+      commitments: [
+        aichi_targets, 
+        sdgs,
+        content['other_targets'] 
+      ]
+    }
+  end
+
+  private
+    def content
```
Levia: Could you please just add an extra line between private and the defitions? Also removing the indentation so it's at the same level as the public methods :)

stacytalbot: It already has that on mine locally?!

stacytalbot: Oh - i was looking in the wrong file haha, too many carols in here

app/controllers/site_controller.rb
Levia: 🤔 🤔 🤔

PR #29, opened 2019-10-22
Sdg

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/15

- Add SDG wheel
- Content to be added in different PR


app/controllers/site_controller.rb
```diff
@@ -12,6 +12,19 @@ def index
...
+    doughnut_chart.each do |item|
+      @doughnut_chart.push({
+        'title': item[:title],
+        'colour': item[:colour],
+        'icon': ActionController::Base.helpers.image_url(item[:icon]),
+        'description': item[:description],
+        'url': item[:url]
+      })
+    end
+
```
WKocur: Could maybe go in a serializer?

stacytalbot: Yea was leaving that for Andrew/Ferdi

PR #33, opened 2019-10-29
Initial work towards the csv parser for habitat data

drewpotter:
This is the initial work towards the CSV parser for the habitat data. Please do the following:
`rails db:migrate db:seed`
`bundle exec rake import:prebakedstats`


db/schema.rb
```diff
@@ -51,20 +67,23 @@
     t.text "wms_url"
   end
 
-  create_table "static_stats", force: :cascade do |t|
+  create_table "regional_stats", force: :cascade do |t|
```
Levia: As this is identical to the `country_stats` (total_points is missing from the country one but we could potentially rename and use the same column both for area and points) we could define just one table for the stats and have also the `countries` table to have both countries and regions; in which case, the new combined table will also have an extra column to distinguish between countries and regions.

Another approach would be to use polymorphic associations if countries and regions are separate tables

drewpotter: Hi Ferdi. I need to completely redo this CSV parser work. As I wrote it before I had the regional data, so I didn't actually know about that data. So potentially all of the PR is wrong.

Levia: Not completely wrong but needs some reworking for sure.
By the way, the comment above was related to what we were discussing earlier about having one table only for geo entities and stats. 
Just thought about making a comment here which might help with that.

app/controllers/site_controller.rb
```diff
@@ -46,33 +46,61 @@ def index
...
+                      geo_entity_stats.habitat_id as habitat_id,
+                      geo_entity_stats.total_value AS total_value
+
+                      FROM geo_entities
+                      INNER JOIN geo_entity_stats ON geo_entities.id = geo_entity_stats.geo_entity_id
+                      INNER JOIN habitats ON habitats.id = geo_entity_stats.habitat_id
+                      WHERE geo_entities.iso3 IS NOT NULL
+                      AND habitat_id = #{@habitat.id}
+                      ORDER BY total_value DESC
+
```
Levia: I know this is not finished, but just a note for the future.
You can use `LIMIT 5` after the `ORDER BY` to fetch the first 5 elements only instead of doing that with ActiveRecord

app/controllers/site_controller.rb
```diff
@@ -46,33 +46,61 @@ def index
...
+             #GROUP BY geo_entities.id, geo_entities.iso3, geo_entities.name, geo_entity_stats.total_value
+
+    geo_entity_stats = ActiveRecord::Base.connection.execute(query)
+
+    puts geo_entity_stats.inspect
+    geo_entity_stats.first(5).each do |geo_entity_stat|
+      puts geo_entity_stat.inspect
+    end
+    byebug
```
Levia: Could you please get rid of the byebug bits before committing when possible?
So other developers can test things out without having to refactor the code and result in uncommitted changes afterwards :)

drewpotter: Apologies for the byebug! I think I made this mistake before in the past. I will remove it and push again

app/controllers/countries_controller.rb
```diff
@@ -3,7 +3,7 @@ def index
   end
 
   def show
-    @country = Country.find(params[:id])
+    @country = GeoEntity.find_by(iso3: params[:id])
```
drewpotter: @Levia I already tried to get iso3 codes being used instead of the id however, we don't know yet what the url should look like, so I think I will revert this change!

drewpotter: Actually, since I changed to using `GeoEntity` to contain both `countries` and `regions` the existing code to link to a particular `id` won't be the same and it will not link to the correct country.

drewpotter: I just fixed that now

app/helpers/application_helper.rb
```diff
@@ -35,7 +35,7 @@ def get_nav_items
       countries: [
         {
           title: I18n.t("countries.#{yml_key}.title"),
-          url: country_path(54)
+          url: country_path(57)
```
Levia: I think I would already start using ISO codes at least as those are better unique keys then ids in this case.
Development environments can break easily because ids can be different depending on how countries have been imported.

Also I can't really see an advantage of using ids even from a user perspective

drewpotter: Sure thing @Levia I will switch to using ISO3 codes in this PR.

lib/tasks/import_prebakedstats.rake
```diff
@@ -0,0 +1,160 @@
...
+        parse_geo_entity_stat csv_headers, csv_new_row, geo_entity_type, habitat
+      end
+    end
+  end
+
+  def parse_change_stat csv_headers, csv_row, geo_entity_type, habitat
+    total_value_years = {}
+ 
+    parse_change_stat_hash = {
+      iso3: csv_headers[0],
```
Levia: if the headers are already following the column names definition, you could use `slice` or the entire hash.
You can check the `import_redlist_data.rake` for reference in general.

lib/tasks/import_prebakedstats.rake
```diff
@@ -0,0 +1,160 @@
...
+      end
+    end
+
+    geo_entity = fetch_geo_entity(name, iso3)
+    insert_change_stat(habitat, geo_entity, total_value_years, protected_value, protected_percentage)
+  end
+
+  def parse_geo_entity_stat csv_headers, csv_row, geo_entity_type, habitat
+    parse_geo_entity_stat_hash = {
+      iso3: csv_headers[0],
```
Levia: same here about the header with slice or using the entire hash as is from the CSV.

lib/tasks/import_prebakedstats.rake
```diff
@@ -0,0 +1,160 @@
...
+      :total_value_2009,
+      :total_value_2010_baseline,
+      :total_value_2015,
+      :total_value_2016
+    ].freeze
+
+    iso3, name = nil
+    protected_value, protected_percentage = 0
+
+    parse_change_stat_hash.keys.each do |key|
```
Levia: I think this logic can be reworked following what I mentioned above.
Again you could use the `import_redlist_data.rake` as reference.

lib/tasks/import_prebakedstats.rake
```diff
@@ -0,0 +1,160 @@
...
+      iso3: csv_headers[0],
+      total_value: csv_headers[1],
+      total_value_protected: csv_headers[2],
+      protected_percentage: csv_headers[3].chomp
+    }.freeze
+
+    iso3, name = nil
+    total_value, total_value_protected, protected_percentage = 0
+
+    parse_geo_entity_stat_hash.keys.each do |key|
```
Levia: Same here.
You could use the `import_redlist_data.rake` as reference.

lib/tasks/import_prebakedstats.rake
```diff
@@ -0,0 +1,160 @@
...
+
+    puts "GeoEntityStat is: #{ges.inspect}"
+    
+  end
+
+  def insert_change_stat(habitat, geo_entity, total_value_years, protected_value, protected_percentage)
+    habitat = Habitat.find_by(name: habitat)
+
+    puts "GeoEntity is: #{geo_entity&.name}, habitat is: #{habitat.name}"
+    cs = ChangeStat.create(habitat: habitat, geo_entity: geo_entity,
```
Levia: Here as well, you could use the assumption that csv headers should already be formatted followin the column names definitions.

app/controllers/countries_controller.rb
```diff
@@ -87,20 +87,39 @@ def show
...
-        title: get_habitat_title('coralreefs'),
-        text: I18n.t('countries.shared.habitat_change.chart_text', km: 20, habitat: get_habitat_title('coralreefs'), years: '2000-2019'),
-        change: -1.4, 
-      },
-      { 
-        id: 'mangroves', 
-        title: get_habitat_title('mangroves'),
-        text: I18n.t('countries.shared.habitat_change.chart_text', km: 40, habitat: get_habitat_title('mangroves'), years: '2000-2019'),
-        change: 3.2
+    # @habitat_change = [
```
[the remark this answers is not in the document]
drewpotter: Apologies! I will fix that.

app/controllers/countries_controller.rb
```diff
@@ -87,20 +87,39 @@ def show
...
+    #     id: 'mangroves', 
+    #     title: get_habitat_title('mangroves'),
+    #     text: I18n.t('countries.shared.habitat_change.chart_text', km: 40, habitat: get_habitat_title('mangroves'), years: '2000-2019'),
+    #     change: 3.2
+    #   }]
+
+    # @habitat_change = @habitat_change.to_json
+
+    @habitat_change = []
+    Habitat.all.each do |habitat|
```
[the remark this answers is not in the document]
drewpotter: Thanks!

app/models/habitat.rb
```diff
@@ -11,9 +11,42 @@ def protected_title
     "Percentage of #{title.downcase} that occur within a marine protected area"
   end
 
-  def calculate_global_coverage
-    coverage = Carto.new(name).total_area.first['sum']
-    update_attributes(global_coverage: coverage.round)
+  def calculate_country_cover_change(iso3)
+    return country_cover_change = {
```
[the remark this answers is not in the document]
drewpotter: Thanks!

app/models/habitat.rb
```diff
@@ -11,9 +11,42 @@ def protected_title
...
+      change_percentage: 0,
+      change_km: 0
+    } unless name == "mangroves"
+    geo_entity_id = GeoEntity.find_by(iso3: iso3)
```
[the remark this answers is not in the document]
drewpotter: Many thanks!

app/models/habitat.rb
```diff
@@ -11,9 +11,42 @@ def protected_title
...
+    habitat_base_year = ChangeStat.where(habitat_id: id, geo_entity_id: geo_entity_id).pluck(:total_value_2010).first
```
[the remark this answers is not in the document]
drewpotter: Many thanks!

app/models/habitat.rb
```diff
@@ -11,9 +11,42 @@ def protected_title
...
+    habitat_last_year = ChangeStat.where(habitat_id: id, geo_entity_id: geo_entity_id).pluck(:total_value_2016).first
```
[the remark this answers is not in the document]
drewpotter: Many thanks!

app/models/habitat.rb
```diff
@@ -11,9 +11,42 @@ def protected_title
...
+
+    change_km = habitat_last_year - habitat_base_year
+    change_percentage = (change_km/habitat_base_year) * 100
+    
+    country_cover_change = {
+      change_km: change_km.round(2),
+      change_percentage: change_percentage.round(2)
+    }
+    country_cover_change
```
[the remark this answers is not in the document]
drewpotter: Cool thanks!

app/models/habitat.rb
```diff
@@ -11,9 +11,42 @@ def protected_title
...
+  end
+
+  def calculate_global_cover_change
+    return global_cover_change = {
```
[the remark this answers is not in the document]
drewpotter: Thanks!

app/models/habitat.rb
```diff
@@ -11,9 +11,42 @@ def protected_title
...
+      change_km: 0,
+      change_percentage: 0,
+      baseline_total: 0
+    } unless name == "mangroves"
+    habitat_base_year = ChangeStat.where(habitat_id: id).pluck(:total_value_2010).inject(0) { |sum, x| sum + x }
```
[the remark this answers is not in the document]
drewpotter: 👍

lib/tasks/import_prebakedstats.rake
```diff
@@ -0,0 +1,160 @@
...
+
+        puts csv_file
+        import_new_csv_file(geo_entity_type, habitat.name, csv_file)
+      end
+    end
+  end
+
+  def import_new_csv_file(geo_entity_type, habitat, csv_file)
+    filename = "#{Rails.root}/lib/data/#{csv_file}"
+    csv = File.open(filename, encoding: "utf-8")
```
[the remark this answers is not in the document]
drewpotter: @Levia I just wanted to clarify what you meant by not needing the line below fetching the headers if I am are using the `headers: true` param in the parse method below.

lib/tasks/import_prebakedstats.rake
```diff
@@ -0,0 +1,160 @@
...
+    csv_headers = File.readlines(csv).first.split(",")
+
+    CSV.parse(csv, headers: true, encoding: "utf-8") do |row|
+      csv_new_row = row.to_hash
+
+      if csv_headers.grep(/baseline/).any?
```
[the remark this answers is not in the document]
drewpotter: @Levia I am determining the type of CSV whether it's a `ChangeStat` based upon whether it has the `baseline` present in the headers. Please advise if that is not correct.

lib/tasks/import_prebakedstats.rake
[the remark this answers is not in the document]
Levia: What I meant is that `headers: true`, within the `parse` or `foreach` method for `CSV`, provides you with the header information for each row in form of a hash.
So, each `row` will be like:
```
  {
    column_name1: value1,
    column_name2: value2,
    baseline_year: baseline_value,
    ...
}
```
The keys of the hash are already representing your header, so you probably wont' need to fetch it again below.

Hope it makes sense

lib/tasks/import_prebakedstats.rake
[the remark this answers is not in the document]
Levia: @drewpotter Please see my comment above as it's strictly related to this as well.
You will already have the `baseline` key within the `row` hash

lib/tasks/import_prebakedstats.rake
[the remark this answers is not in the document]
drewpotter: Oh thanks @Levia !

lib/tasks/import_prebakedstats.rake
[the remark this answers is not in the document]
drewpotter: Many thanks!

drewpotter:
Many thanks Ferdi! I think I have addressed every issue now 🎉

lib/tasks/import_prebakedstats.rake
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  def parse_geo_entity_stat csv_row, geo_entity_type, habitat
+    geo_entity = fetch_geo_entity(csv_row["name"] | csv_row["Region"], csv_row["iso3"])
+    insert_geo_entity_stat(habitat, geo_entity, csv_row)
+  end
+
+  def insert_geo_entity_stat(habitat, geo_entity, csv_row)
+    habitat = Habitat.find_by(name: habitat)
+    ges = GeoEntityStat.create(habitat: habitat, geo_entity: geo_entity, protected_value: csv_row["total_value_protected"]&.strip || 0,
```
Levia: Two things here:

1. As the `ges` variable is not used anywhere, you can get rid of the assignment as the value on the right hand side will be returned anyway
2. I would probably put the computed `protected_value` into a variable first to clean up the code a bit.
Same below for `total_value` and `protected_percentage`.

lib/tasks/import_prebakedstats.rake
```diff
@@ -0,0 +1,73 @@
...
+      if row.select {|k,v| /baseline/i =~ k}.any?
+        parse_change_stat row, geo_entity_type, habitat
+      else
+        parse_geo_entity_stat row, geo_entity_type, habitat
+      end
+    end
+  end
+
+  def parse_change_stat csv_row, geo_entity_type, habitat
+    geo_entity = fetch_geo_entity(csv_row["name"] | csv_row["Region"], csv_row["iso3"])
```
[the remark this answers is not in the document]
Levia: I think you meant `||`.
Single `|` is a bitwise operator which is not achieving the intended behaviour here I think.
For example, if one between `csv_row["name"]` and `csv_row["Region"]` is populated you will get `true`, `false` otherwise, but the first param for `fetch_geo_entity` is expecting a name.

app/models/change_stat.rb
```diff
@@ -1,4 +1,4 @@
 class ChangeStat < ApplicationRecord
   belongs_to :habitat, -> { where(name: 'mangroves') }
-  belongs_to :country
+  belongs_to :geo_entity
```
Levia: Below here, I would also add a validation for the uniqueness of `habitat_id` and `geo_entity_id`. This is to make sure we have no duplicates, also when importing using the rake task.

app/serializers/habitat_cover_serializer.rb
```diff
@@ -9,16 +9,19 @@ def serialize
...
+      habitat_global_change = Habitat.find_by(name: habitat.to_s).calculate_global_cover_change
+
       habitat_cover_item[:text] =
         I18n.t(
           "home.habitat_cover.habitats.#{habitat}.text",
-          change_km: -80000,
-          change_percentage: -8,
+          change_km: habitat_global_change[:change_km],
+          change_percentage: habitat_global_change[:change_percentage],
           baseline_year: 999999,
```
Levia: Is this meant to be populated now?
Same for `original_total` below?

lib/tasks/import_prebakedstats.rake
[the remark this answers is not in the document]
drewpotter: Thanks!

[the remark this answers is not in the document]
drewpotter: @Levia I changed to `||` but I don't get the correct statistics post import. I will investigate further why this is happening.

drewpotter:
@Levia I think I have addressed every issue in this PR. Although I am getting incorrect data from the importer for the global coverage, however the individual country data is correct for Costa Rica.

PR #34, opened 2019-10-29
Add change data for mangroves

Levia:
## Description

* Add change stats table to database. Currently only used for mangroves habitat
* Update `countries.csv` to include a "disputed territories" record (need `rake db:seed`)
* Add relevant associations to models, making sure only mangroves will have change stats data

### DIFF
--- a/app/models/country.rb
+++ b/app/models/country.rb
@@ -1,5 +1,5 @@
 class Country < ApplicationRecord
-  has_many :static_stat
+  has_many :static_stats
   # At the moment, only mangroves have got change stats,
   # which means there can only be one change_stat record per country.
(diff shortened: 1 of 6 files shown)


app/models/country.rb
```diff
@@ -1,2 +1,7 @@
 class Country < ApplicationRecord
+  has_many :static_stat
```
drewpotter: I noticed we have `has_many static_stat` in Country and `has_many static_stats` in Habitat. I just wondered about the pluralisation both with and without. Should they both be the same as each other?

drewpotter:
Happy to merge this! Just one query about the pluralisation of the association mentioned.

PR #36, opened 2019-10-31
Red list country

stacytalbot:
Merge after https://github.com/unepwcmc/ocean-plus-habitats/pull/35

- Refactor the table so that it works with a differing number of habitats.
- Match design on the country page


WKocur:
Is the total row meant to show on both the global and country pages? @stacytalbot

PR #39, opened 2019-11-04
Chart habitat change

stacytalbot:


app/controllers/countries_controller.rb
```diff
@@ -29,9 +28,36 @@ def show
...
+  private
+  def habitats
+    I18n.t('global.habitats')
   end
 
   def getStatusText status
     I18n.t("countries.shared.habitats_present.title_#{status}")
   end
+
+  def getHabitatTitle id
```
WKocur: Probably should keep to snake_case for rb

stacytalbot: My bad i always do this by mistake

config/deploy/staging.rb
```diff
@@ -1,5 +1,5 @@
 set :stage, :staging
-set :branch, "supported-staging"
+set :branch, "refresh"
```
WKocur: ?!

stacytalbot: :/

app/controllers/countries_controller.rb
stacytalbot: done

PR #45, opened 2019-11-11
Component/map

WKocur:
Adds the map without the download functionality yet.


WKocur:
Sorry, I should have separated out the linter stuff and fixes...

config/locales/shared/en.yml
```diff
@@ -6,4 +6,6 @@ en:
     footer_downloads_title: Download datasets
 
     homepage: "%{name} home page"
-    visit: Visit %{name}
\ No newline at end of file
+    visit: Visit %{name}
+
```
stacytalbot: @WKocur whats with the blank line in this file

WKocur: No particular reason, can be next to visit if you want

app/views/countries/show.html.erb
```diff
@@ -1,6 +1,7 @@
-<%= render '/global/header-country', locales: { title: @country.name }%>
+<%= render '/global/header-country' %>
 
 <%= render '/countries/partials/habitats-present' %>
+<%= render partial: '/countries/partials/map', locals: { iso3: @country.iso3 } %>
```
stacytalbot: You don't need to do this - @country.iso3 is available in the partial

app/views/countries/partials/_map.html.erb
```diff
@@ -0,0 +1,6 @@
+<section class="container section-padding--bottom">
+  <v-map-interactive iso3="<%= iso3 %>"></v-map-interactive>
```
stacytalbot: Can change iso3 to @country.iso3

WKocur: I thought it was bad practice to use instance variables in partials, so have assigned a local variable in the parent view. Happy to change if I'm wrong on this

WKocur: (Presumably because the partial is then connected to the controller and thus not so reusable as before?)

stacytalbot: Ah ok yea - leave as is!

PR #46, opened 2019-11-12
SDG Wheel - sources

stacytalbot:
Add all sources to SDG wheel


WKocur:
@stacytalbot conflicts

PR #48, opened 2019-11-13
Component/map button

WKocur:


app/assets/stylesheets/resources/variables/_base.scss
```diff
@@ -11,7 +11,11 @@ $black: #000000;
 
 $green: #22A37B;
 
+$blue-light: #0AD4FF;
+$blue: #0895B4;
+
 $navy: #0A2F3C; 
+$navy-2: #18424E;
```
stacytalbot: Either this can replace one of the existing navys or they all need looking at - at the end as some of these might not be used anymore/can be merged

app/javascript/components/map/filters/MapDownloadButton.vue
```diff
@@ -0,0 +1,97 @@
...
+      isActive: false,
+      contentId: `map-download-button-content-${this._uid}`,
+      options: [
+        {
+          label: 'CSV',
+          id: 'csv',
+          link: ''
+        },
+        {
+          label: 'SHP',
```
stacytalbot: Should the label names be hardcoded here or passed in as options?

app/views/layouts/application.html.erb
```diff
@@ -6,6 +6,10 @@
 
   <body>
     <div id="v-app" v-cloak>
+      <translator 
+        :translations="<%= @current_translations.to_json %>"
+        :locale="<%= @current_language.to_json %>">
+      </translator>
```
stacytalbot: Why has this been added? We don't have anything in the designs

WKocur: This is the thing that adds the translations to a Vue store. It doesn't have to be a component tbh. I could get the rails json in application and assign to a window variable and just access that... Maybe that is better? window.translations?

app/javascript/components/map/filters/MapDownloadButton.vue
WKocur: They will be passed in, just haven't mapped back to the controller yet.

app/assets/stylesheets/resources/variables/_base.scss
WKocur: Agreed

PR #50, opened 2019-11-14
Scroll to top

WKocur:
Scroll to top button


app/javascript/components/scroll-to-top/ScrollToTop.vue
```diff
@@ -0,0 +1,51 @@
...
+      }
+    },
+
+    // t = current time, b = start value, c = change in value, d = duration
+    easeInOutQuad (t, b, c, d) {
+      t /= d / 2
+      if (t < 1) { return c / 2 * t * t + b }
+      t--
+
+      return -c / 2 * (t * (t - 2) - 1) + b
```
stacytalbot: haha where did this come from!

WKocur: T'internet ;) I think it's just suvat though

WKocur:
@stacytalbot How much faster - can do. Will extract variable and call it speed

PR #51, opened 2019-11-15
Nav tertiary

stacytalbot:
Add Tertiary nav - excluding the last 2 links because the sections haven't been created yet.
Haven't updated styling as there is a chance we may build the other sections before the site goes live.


WKocur:
This is fine, but we do now have two different methods for scrolling. Would make sense to use the same for both. I don't really mind which. We can leave for now though if time is short

PR #64, opened 2019-12-10
Frontend tweaks

WKocur:
https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/51
https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/52


app/helpers/application_helper.rb
```diff
@@ -52,4 +52,12 @@ def get_habitat_icon_class(habitat, status='')
 
     "icon--#{habitat}#{status}"
   end
+
+  def country_path (country)
+    '/countries/' + country.gsub(/ /, '-').gsub("'", '%27').downcase
```
stacytalbot: This url isn't going to make sense for the regional pages

WKocur: @stacytalbot  Good point - could switch to regions instead of countries - will check with roger

WKocur:
> Getting `ZeroDivisionError in SiteController#index` error

Is this relevant to this PR? Not sure if you might need to do some new import... Maybe @Levia will know

app/helpers/application_helper.rb
stacytalbot: or leave it out all together?

WKocur: Also a possibility yeah. I've been too railsized ;)

rogeringle: What do you need to check with me @wkocur?

WKocur: What are you doing here!? ;) there's a comment on the codebase ticket https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/52

rogeringle: I just saw my name in the Slack feed! 😆 

Have answered on Codebase...

WKocur: @stacytalbot done

PR #65, opened 2019-12-10
add dynamic footer citation

WKocur:
https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/53
https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/54


WKocur:
@stacytalbot done

PR #66, opened 2019-12-13
Redlist data

Levia:
## Description

* Populate IUCN categories charts with real species data.
* Refactor and clean up the code a bit using helpers more.

## Notes
Codebase ticket [here](https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/39)

### DIFF
--- a/app/models/species.rb
+++ b/app/models/species.rb
@@ -26,14 +26,14 @@ class Species < ApplicationRecord
   def self.count_species(species = nil)
     species ||= all
     hash = count_by_category_and_habitat(species).compact
-    hash.each do |habitat|
-      hash[habitat.first] = Species.fill_and_sort_by_category(habitat.last)
+    hash.each do |habitat, categories_data|
+      hash[habitat] = Species.fill_and_sort_by_category(categories_data)
     end
   end
 
   def self.count_by_category_and_habitat(species)
     groupings = {}
-    species.group_by { |s| [s.redlist_status, s.habitat.name] }.map do |key, values|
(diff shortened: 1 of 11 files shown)


app/models/species.rb
```diff
@@ -20,4 +22,31 @@ class Species < ApplicationRecord
   scope :near_threatened, -> { where(redlist_status: NEAR_THREATENED) }
   scope :threatened_and_near, -> { where(redlist_status: [THREATENED, NEAR_THREATENED].flatten) }
   scope :not_threatened, -> { where.not(redlist_status: THREATENED) }
+
+  def self.count_species(species = nil)
+    species ||= all
+    hash = count_by_category_and_habitat(species).compact
+    hash.each do |habitat|
```
WKocur: Is it possible to use a |key, value| (perhaps something like |habitat_id, count_by_category| or similar) syntax here rather than habitat.first and habitat.last? The code like this makes hash look like an array of arrays. Maybe this is just me not being very familiar with ruby conventions!

app/models/species.rb
```diff
@@ -20,4 +22,31 @@ class Species < ApplicationRecord
...
+      hash[habitat.first] = Species.fill_and_sort_by_category(habitat.last)
+    end
+  end
+
+  def self.count_by_category_and_habitat(species)
+    groupings = {}
+    species.group_by { |s| [s.redlist_status, s.habitat.name] }.map do |key, values|
```
WKocur: Should this be .each instead of .map?

Levia:
Thanks @WKocur !
Comments addressed

PR #69, opened 2019-12-20
Habitats present

Levia:
# Description

* Calculate country/regional/global level protection stats
* Add dynamic stats to map filter pane
* Import habitat occurrences into the database
* Show correct occurrences (presence cards)


WKocur:
@Levia Looks good!

PR #70, opened 2020-01-06
Habitats cover tweaks

WKocur:
To merge after habitats-present.

Show habitat change row charts for all habitats that are confirmed as present. If data is not available an empty bar chart is shown.


WKocur:
Screenshot 2020-01-06 at 09 56 49

PR #71, opened 2020-01-06
Targets content

WKocur:
This copies over the targets content from the current live site. Habitat specific other targets are grouped under headings. Styling and content may change in the future.


WKocur:
Screenshot 2020-01-06 at 15 49 48

app/serializers/serializers/targets_serializer.rb
```diff
@@ -0,0 +1,16 @@
+class Serializers::TargetsSerializer < Serializers::Base
```
Levia: This looks more like an helper than a serializer as there are no input arguments and it's not fetching any of the data from the database.
Morover, it's actually using helpers method already; so I would move the entire thing into the helper of the controller that's using it, so this can be used in the view straight away.  

What do you think? @WKocur

WKocur: Makes sense - I was unsure about this already, so will move to a helper.

PR #72, opened 2020-01-07
handle custom bounding boxes for regions

WKocur:
Allows for custom bounding boxes to be configured in yml files for the region/country pages. Waiting on bounding boxes from Osgur, but this can still be merged.


config/locales/countries/mediterranean/en.yml
```diff
@@ -2,6 +2,13 @@ en:
   countries:
     mediterranean:
       title: Mediterranean
+      bounding_box:
```
Levia: This doesn't feel like it should live in a locale file as it's the same across languages and not related to translations :) 
I would put it into a separate yml file related to countries data if we want to stick with ymls.

However, I would be tempted to say that this should be a database attribute for the GeoEntity model; so something to be imported with other data when seeding the database.

WKocur: I think you're right that it belongs as a db attribute. Looks like it would maybe need a new csv? Would this be something you could set up quickly? If not I can give it a go and get your review.

Levia: I can quickly set it up on this PR :) no problem!

WKocur: Thank you!

Levia:
@WKocur This is ready for you to have a look now.
You would need to launch the following from your terminal:

```
> bundle exec rake db:migrate
> bundle exec rake import:bounding_boxes
```

Or alternatively to the last command, you could do:

```
> bundle exec rake import:countries
> bundle exec rake import:regions
```

WKocur:
@Levia I've updated the csv column so that they are all `0 0;0 0` aside from wider caribbean and mediterranean and have added a check for those that are `0 0; 0 0` which replaces this with `[]` before passing to the frontend. 

UAE and indonesia use the bounding boxes from querying a layer provided by osgur and the regions use the values from the csv.

Levia:
@WKocur That makes sense.
However, since the bounding_box attribute defaults already to `[]`, I was wondering that might be better to just leave the related CSV cell empty and amend the importers slightly.

Having all 0s in the database and intepreting them as "no bounding box" could be a bit confusing I believe.

What do you think?

WKocur:
@Levia Completely agree - I hadn't realised that it defaults to []. Will fix that now then :)

Levia:
@WKocur Cool :) Thanks!

WKocur:
@Levia I can't see the code where this defaults - if I leave bounding box blank I get `NoMethodError: undefined method `split' for nil:NilClass` when running the import task

Levia:
@WKocur Yeah you need to amend the import task slightly. I can do that if that's ok for you.
The default attribute is in the migration at the very moment of generating the new attribute on the db table

WKocur:
@Levia Ah ok, I've edited the csvs, so will commit them. If you could amend the import task that would be perfect. Thanks :)

Levia:
Thanks @WKocur .
I've updated the importers.
I would suggest to do a rollback and then run the task again:

```
> bundle exec rake db:rollback
> bundle exec rake import:bounding_boxes
```

WKocur:
@Levia Looks good to me - happy to merge if you are!

PR #78, opened 2020-03-16
Check integrity of redlist api link in the csv

lucacug:


lib/tasks/import_redlist_data.rake
```diff
@@ -35,6 +35,7 @@ namespace :import do
     CSV.foreach(species_filename, headers: true) do |row|
       next unless row['species_id']
       next if Species.find_by(species_id: row['species_id'], habitat_id: habitat.id).present?
+      row['url'] = 'https://'.concat(row['url']) if (row['url'] != 'N/A' && !row['url'].include?('https'))
```
Levia: This should work fine already, however, I would be more precise about the inclusion of `https`.
Even though it won't happen, at the moment, any url containing `https` at any point of the string will trigger that change, so I would do something like:

```
... && !row['url'].match(/^https/)
```

In regexp, `^` means 'starting with'

PR #80, opened 2020-03-18
Ticket 85: Refresh narative menus

nym1k:
Ticket: https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/85

### DIFF
--- a/app/assets/stylesheets/styles/components/_nav.scss
+++ b/app/assets/stylesheets/styles/components/_nav.scss
@@ -2,7 +2,7 @@
 // classes
 //--------------------------------------------------
 .nav {
-  
+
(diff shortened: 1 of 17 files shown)


Levia:
@nym1k  Code-wise it looks good!
However, I noticed some differences between the implementation and the requirements specified in the Codebase ticket.

For example, I can't see `Habitat condition` in the requirements for the global page, while a requirement was to replace `Importance of habitats` with `Importance to wellbeing`, which is currently missing. So in general there are some nav items which shouldn't be there or are not referenced to an item in the page.

Not sure if I missed some conversations about this or if the ticket needs updating.
Hope it makes sense

PR #81, opened 2020-03-18
Ticket 76: Refresh protection dials

nym1k:
Ticket: https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/76

### DIFF
--- a/app/assets/stylesheets/styles/components/cards/_cards-protection-dials.scss
+++ b/app/assets/stylesheets/styles/components/cards/_cards-protection-dials.scss
@@ -1,0 +1,54 @@
+&--protection-dials {
+  @include flex;
+  @include flex-h-between;
+  margin-top: rem-calc(80);
+  margin-bottom: rem-calc(50);
+
+  .card {
+    padding: rem-calc(0 10);
+    width: 20%;
+
+    flex: 1 1 20%;
+
+    &--unknown {
+      .card__dial-circle--colour {
+        display: none;
+      }
+      .card__habitat-status {
+        opacity: .4;
+      }
+    }
+
+      &__status {
+        @include flex;
+        @include flex-v-center;
+        margin-bottom: rem-calc(22);
+      }
+
+        &__dial {
+          margin-bottom: rem-calc(10);
+        }
+
+        &__habitat {
+          text-align: center;
+        }
+
+          &__habitat-title {
+            @include text-protection-dials-title;
+            margin-bottom: rem-calc(15);
+          }
+
+          &__habitat-status {
+            @include text-protection-dials-status;
+
+            display: block;
+          }
+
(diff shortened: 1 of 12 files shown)


app/assets/stylesheets/styles/components/cards/_cards-protection-dials.scss
```diff
@@ -0,0 +1,54 @@
...
+            display: block;
+          }
+
+          &__habitat-percentage {
+            color: $grey;
+            font-size: rem-calc(22);
+
+            display: block;
+          }
+  }
```
Levia: I think there's an indentation issue here. These elements don't seem to be nested into one another so it would probably be better to remove those extra white spaces where possible

app/views/countries/partials/_protection-dials.erb
```diff
@@ -0,0 +1,51 @@
...
+
+  <div class="cards--protection-dials">
+    <% map_datasets.each do |dataset| %>
+      <%
+        dataset_status = @country.occurrences[dataset[:id]]
+        dataset_status_title = I18n.t("countries.shared.habitats_present.title_#{@country.occurrences[dataset[:id]]}")
+        icon_class = get_habitat_icon_class(dataset[:id], @country.occurrences[dataset[:id]])
+        protected_percentage = @country.protection_stats[dataset[:id]] ? @country.protection_stats[dataset[:id]]['protected_percentage'].round(2) : 0
+        total_value = @country.protection_stats[dataset[:id]] ? @country.protection_stats[dataset[:id]]['total_value'].to_i : 0
+        total_units = dataset[:id] == 'coldcorals' ? 'observations' : 'km<sup>2</sup>'.html_safe
```
Levia: I would avoid assigning variables in the view as much as possible. So perhaps those could be moved into the `countries_helper` which already exists in `app/helpers/`.

Some of those could also be reused like embedding the `dataset_status` within the `dataset_status_title`

app/views/countries/partials/_protection-dials.erb
```diff
@@ -0,0 +1,51 @@
+<section class="container section-padding--bottom">
```
Levia: Views file format for rails should conventionally be `.html.erb` instead of just `.erb`

app/assets/stylesheets/styles/components/cards/_cards-protection-dials.scss
```diff
@@ -0,0 +1,46 @@
...
+          display: block;
+        }
+
+          &__habitat-percentage {
+            color: $grey;
+            font-size: rem-calc(22);
+
+            display: block;
+          }
+
```
Levia: I think the indentation here is still off.
I would put all rules at the same indentaion level if they are not nested

PR #83, opened 2020-03-20
Add coastal chart dynamic data

lucacug:
You might need to run the coastal importer to import the corrispective data.
`bundle exec rake import:coastalstats`


lib/tasks/import_coastalstats.rake
```diff
@@ -0,0 +1,47 @@
...
+    import_csv_file(csv_file)
+  end
+
+  def import_csv_file(csv_file)
+    CSV.foreach(csv_file, headers: true, encoding: 'utf-8') do |row|
+      # temporary workaround to catch only one buffer
+      next unless row['buffer'] == '30'
+      habitats = Habitat.all
+      name = row['iso3'] || row['name']
+      geo_entity = fetch_geo_entity(name)
```
Levia: I'm getting an error when importing which is probably related to this method conflicting with the one in the prebakedstats import.
If you need the implementation of this to be different I would suggest to rename it

app/serializers/serializers/representation_habitats_serializer.rb
```diff
@@ -0,0 +1,32 @@
+class Serializers::RepresentationHabitatsSerializer < Serializers::Base
+
+  def initialize(coastal_stat, habitat_stats)
+    super(coastal_stat, 'coastal_stat')
+    super(habitat_stats, 'habitat_stats')
+  end
+
+  def serialize
+    coastal = @coastal_stat.attributes.slice('total_coast_length', 'multiple_habitat_length')
```
Levia: This currently breaks for regions pages as we don't have a coastal_stat object related to those

lucacug:
thanks @Levia . Comments addressed.

PR #84, opened 2020-03-23
Ticket 75 - Refresh new megamenu (Desktop)

nym1k:
Ticket: https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/75


app/javascript/helpers/countries-helpers.js
```diff
@@ -0,0 +1,113 @@
+export const UNDEFINED_ID = '__UNDEFINED__'
```
Levia: I believe this is the only thing currently used from this file.
I would then just move it to the component where this is used and remove this file completely.

app/javascript/mixins/mixin-select-shared.js
```diff
@@ -0,0 +1,205 @@
...
+      this.resetSearchTerm()
+      this.focusSearch()
+    },
+
+    resetSearchTerm () {
+      this.searchTerm = ''
+    },
+
+    scrollToHighlightedOption () {
+      this.$el.querySelector('#' + this.highlightedOptionId).scrollIntoView()
```
Levia: When trying to use up and down arrow keys to navigate the dropdown items I get the following error related to this line: ```Error in callback for watcher "highlightedOptionId": "TypeError: this.$el.querySelector(...) is null"```

Not sure if related, but dropdown items are not highlighted when navigating using the arrow keys; however, I can see it's actually selecting those anyway after hitting enter.

nym1k: @WKocur has resolved this issue

app/javascript/mixins/mixin-select-shared.js
```diff
@@ -0,0 +1,213 @@
...
+    showResetIcon () {
+      return this.searchTerm && this.isActive
+    },
+
+    showSearchIcon () {
+      return this.hasSearchIcon && !this.showResetIcon
+    },
+
+    container () {
+      console.log('hree')
```
stacytalbot, line 66: @nym1k delete console log

WKocur: Oops, this was me 🐙

PR #85, opened 2020-03-23
Refresh new data import

lucacug:
This includes the refactor of the importers (prebaked and redlist rake task) based on the new csv files provided.


lib/tasks/import_new_redlist_data.rake
```diff
@@ -0,0 +1,158 @@
...
+
+    row['url'] = 'https://'.concat(row['url']) if (row['url'] != 'N/A' && !row['url'].match(/^https/))
+    row_hash = row.to_hash.slice(*species_header)
+    species = Species.new(row_hash.merge(habitat_id: habitat.id))
+    report_failed(row, species.errors.messages) unless species.save
+  end
+
+  def import_countries_species(row, habitat)
+    iso3 = row['iso3']
+
```
lucacug: I would add a check here like
 `return if scientific_name == 'NA'` just to avoid having many rows in the failed report like 
```ABW,Cold-water corals,NA,NA,NA,NA,NA,1,NA,"Species, habitat pair not found!"```

File: lib/tasks/import_new_redlist_data.rake [hunk 2]
```diff
@@ -0,0 +1,158 @@
...
+  end
+
+  def import_species(row, habitat)
+    species_header = Species.new.attributes.keys.freeze
+    scientific_name = row['scientific_name']
+
+    return if scientific_name == 'NA'
+    return if Species.find_by(scientific_name: scientific_name, habitat_id: habitat.id).present?
+
+    row['url'] = 'https://'.concat(row['url']) if (row['url'] != 'N/A' && !row['url'].match(/^https/))
```
lucacug: I think `N/A` is something I missed during the refactor; that should be `NA` now. Sorry, my bad.

lib/tasks/import_new_redlist_data.rake
```diff
@@ -0,0 +1,158 @@
...
+    true
+  end
+
+  def report_failed(row, error)
+    @failed_report << row.headers.map(&:to_s).push('error') if @failed_report.empty?
+    @failed_report << row.values_at(*row.headers).push(error)
+  end
+
+  def log_failed(geo_type)
+    if @failed_report.present?
```
lucacug: just a very minor thing here, I would do `return unless @failed_report.present?` rather than including everything into the if block

lib/tasks/import_new_redlist_data.rake
Levia: My original intention was to report every single row that failed, so it's easier to understand what happened also by counting the number of records successfully imported and the ones that failed. 
For example, I spent quite a bit of time understanding why the number of imported ones was different; with such a report it was easier.

Hope that makes sense

Levia:
@lucacug Thanks! Comments addressed with the exception of skipping the 'NA' species for the report

PR #86, opened 2020-05-04
removed text from side-panel and redundant method from serializer

stanleypliu:
removed text from side-panel of all maps, along with the method from the serializer which populates the dataset description. Have not yet styled the resulting list to add more padding between list items.

### DIFF
--- a/app/assets/stylesheets/styles/components/map/_map.scss
+++ b/app/assets/stylesheets/styles/components/map/_map.scss
@@ -39,6 +39,9 @@
 
   &__list {
     @include ul-unstyled;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/styles/components/map/_map.scss
```diff
@@ -39,6 +39,9 @@
 
   &__list {
     @include ul-unstyled;
+    li {
```
stacytalbot, line 42: rather than nest the li here it would be better to add a class to them like .map-filters__list-li

PR #88, opened 2020-05-04
Refresh eez map

lucacug:


lucacug:
The eez map is working for me! I just wonder if the habitat order on the filter pane should be the same of the map above. Maybe one for @rogeringle 
Screenshot from 2020-05-13 11-05-43
Screenshot from 2020-05-13 11-05-38

lucacug:
@stanleypliu Happy for this to be merged after conflicts solved

PR #90, opened 2020-05-05
Changes to redlist

stanleypliu:
https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/74 - addressed all tickets


app/views/global/_head.html.erb
```diff
@@ -37,6 +38,10 @@
 <script src='https://api.tiles.mapbox.com/mapbox-gl-js/v0.47.0/mapbox-gl.js'></script>
 <script src='https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-draw/v1.0.4/mapbox-gl-draw.js'></script>
 
+<!-- Font Awesome icons -->
+<link rel="stylesheet" href="https://use.fontawesome.com/releases/v5.1.0/css/all.css">
```
stacytalbot, line 42: We are trying to avoid using icon fonts - you should be able to get SVGs from the design - or ask charlotte for it.

app/javascript/components/content/ExampleSpecies.vue
```diff
@@ -3,32 +3,36 @@
...
-          <a 
-            :href="example.redlist_url" 
+
+          <a
+            :href="example.redlist_url"
             class="card__link"
+            target="_blank"
           >
             IUCN Species page
+          <i class="fas fa-external-link-square-alt"></i>
```
stacytalbot, line 35: replace font awesome with SVG file

app/views/shared/_red-list-table.html.erb
```diff
@@ -13,17 +13,36 @@
   </div>
 
   <% @red_list_data.each do |habitat| %>
-    <% if habitat['data'].present? %>
       <div class="table__column">
         <div class="table__cell">
-          <i class="table__icon <%= get_habitat_icon_class(habitat[:id]) %>" alt="<%= habitat[:title] %> icon"></i>
+          <% if !habitat['data'].present? %>
+            <i class="table__icon <%= get_habitat_icon_class(habitat[:id])-%>-absent" alt="<%= habitat[:title] %> icon"></i>
```
stacytalbot, line 19: Looks like the css needs updating so that the absent icon uses one which is a paler grey (the one used higher up the page.)

app/javascript/components/content/ExampleSpecies.vue
```diff
@@ -3,32 +3,36 @@
...
+          :src="example.image"
+          :alt="example.name_common"
           class="card__image"
+          :class="{'card__image-svg': example.has_image == false}"
         >
 
         <div class="card__content">
           <p class="card__title">
             {{ example.name_common }}
+            <br v-if="!example.name_common">
```
stacytalbot, line 23: I would change this so that you apply some margin to create the gap rather than add in a <br>. Maybe put a span around the common name.

app/javascript/components/content/ExampleSpecies.vue
```diff
@@ -3,32 +3,36 @@
     <h3 class="cards__title">
       {{ content.title }}
     </h3>
-    
+    <p v-if="!examples">No species could be found for this particular habitat and category.</p>
```
stacytalbot, line 6: Since this is being updated every time a new species is selected it is more performant to use `v-show` here. Only use `v-if` for things that are set once on page load.

app/views/shared/_red-list-table.html.erb
```diff
@@ -13,17 +13,36 @@
   </div>
 
   <% @red_list_data.each do |habitat| %>
-    <% if habitat['data'].present? %>
       <div class="table__column">
         <div class="table__cell">
-          <i class="table__icon <%= get_habitat_icon_class(habitat[:id]) %>" alt="<%= habitat[:title] %> icon"></i>
+          <% if !habitat['data'].present? %>
```
stacytalbot, line 18: Can you see if this `if` can be pulled out here and added to the `get_habitat_icon_class` so that you don't need to duplicate the html in the view.

app/javascript/components/content/ExampleSpecies.vue
```diff
@@ -3,32 +3,36 @@
...
     <div class="cards__cards">
-      <div 
+      <div
```
stacytalbot, line 8: Can you just add some padding somewhere here or decrease the image size so that they aren't so close together - see design: https://xd.adobe.com/spec/f60a4b5b-8b02-4d24-43c8-137742d2ec01-803b/screen/fe06cc73-2d92-4358-a37c-53e3b1ea2f2a/Country-D-

app/javascript/components/content/ExampleSpecies.vue
stanleypliu: I've added a margin-top to the <p> element if there is no common name.

app/javascript/components/content/ExampleSpecies.vue
stanleypliu: Have decreased the image size slightly, but on my screen there is space between them.

PR #91, opened 2020-05-06
Ticket 73

stanleypliu:
Change in habitat graph for country and region pages now the same format as the global page. Confirmed absence, presences and unknown statuses correctly handled, as well as no-data situations.


stanleypliu:
Just realised I need to continue working on the _habitat-change partial to put it in /shared and make it usable by both the homepage and each country page as at the moment, it is very similiar.

PR #92, opened 2020-05-11
Country map fix

stanleypliu:
This branch fixes the country habitat map not showing when the EEZ map was implemented, as well as carrying out almost all the necessary styling to replicate the appearance of Adam's design [here](https://unep-wcmc.codebasehq.com/upload/40b4cf21-b334-5060-7c1e-992dad6250e7/show/original) 

image

NB: At the time of the screenshot the servers hosting the map tile data were broken hence the select filter function does not work.

@lucacug Could you check that the map looks as intended?


app/assets/stylesheets/resources/variables/_base.scss
```diff
@@ -37,6 +37,9 @@ $color-mangroves: #D6A621;
 $color-seagrasses: $theme-blue;
 $color-coldcorals: $theme-red;
 
+$checkbox-checked: #1AD866;
+$eezmap-filter-active: #07202A;
```
stacytalbot, line 41: Rather than creating new colours try and use the existing ones so that we don't end up with a huge list. Ideally designers shouldn't be adding new colours that are only slightly different to the existing colour palette. It looks like you can use navy-xdark here for one of them. For the green one set a new green variable to that hex - then you can reference the green variable - this way it is clear what colour each variable is. Not many people can read hex codes ;)

app/views/countries/partials/_map.html.erb
```diff
@@ -1,6 +1,7 @@
 <section class="container section-padding--top">
   <v-map-interactive
     iso3="<%= iso3 %>"
+    id="'map-target'"
```
stacytalbot, line 4: Does the id here need to be in single quotes?

app/javascript/components/map/VMapInteractive.vue
```diff
@@ -7,13 +7,32 @@
...
+        <div
+          v-for="extent in datasets[0].sourceLayers"
+          :key="extent.id"
+          class="eez-map-legend_keys"
+        >
+          <span
+            :key="extent.id"
+            class="eez-map-legend_key"
+            :style="{ 'background-color': extent.color }"
+          /><p>{{ extent.sub_name | correct }}</p>
```
stacytalbot, line 32: I would suggest changing this `p` to a `span` and aligning them vertically with flex to match the design.

app/javascript/components/map/filters/FilterPane.vue
```diff
@@ -1,10 +1,16 @@
 <template>
   <div class="map-filters flex flex-column">
+    <p
+      v-show="isEez"
+      class="map-filters__eezMessage"
```
stacytalbot, line 5: I would say this title looks like it should be left aligned not centered.

app/javascript/components/map/filters/FilterPane.vue
```diff
@@ -42,16 +43,18 @@ export default {
...
-      type: Boolean,
-      default: false
     }
   },
 
   data () {
     return {
-      isActive: true
+      isActive: true,
+      eezMessage: 'Select an Ocean Habitat'
```
stacytalbot, line 52: Avoid hardcoding content in the vue files - move this to a yml file and pass in as a config prop.

app/assets/stylesheets/styles/components/map/_map.scss
```diff
@@ -177,3 +210,33 @@
...
+
+.eez-map-legend {
+  font-size: 0.7rem;
+  height: 11%;
+  width: 100%;
+  display: flex;
+  justify-content: space-around;
+  align-items: center;
+
+  &_keys {
```
stacytalbot, line 226: In bem for elements that are nested you need to use two underscores `&__keys`

app/assets/stylesheets/styles/components/map/_map.scss
```diff
@@ -177,3 +210,33 @@
...
+    @include flex-no-shrink;
+    display: flex;
+  }
+
+  &_key {
```
stacytalbot, line 231: Correct the bem syntax here too

app/assets/stylesheets/styles/components/map/_map.scss
```diff
@@ -8,12 +8,29 @@
 
     flex-direction: row-reverse;
     position: relative;
+
+    border: 1px solid $navy;
+  }
+}
+
+.maptype {
+  &_eezmap {
```
stacytalbot, line 17: I think you are using modifiers here? So eezmap and habitatmap are two different maptypes - in which case you need two hyphens like so `&--eezmap` and `&--habitatmap`

app/javascript/components/map/VMapInteractive.vue
```diff
@@ -7,13 +7,32 @@
...
+        :search="search"
+        :allow-no-selected-dataset="allowNoSelectedDataset"
+        :mapbox-token="mapboxToken"
+        :bounding-box="initBoundingBox"
+      />
+      <div
+        v-if="isEez"
+        class="eez-map-legend"
+      >
+        <p><i>Extent of marine protected area coverage (%):</i></p>
```
stacytalbot, line 22: It might be quite nice to pull the legend out into a new vue component - we have a few that have been used on other projects.

stacytalbot: Maybe something similar to this https://github.com/unepwcmc/pp-live-report/blob/master/app/javascript/components/charts/ChartLegend.vue

app/javascript/components/map/filters/Dataset.vue
```diff
@@ -8,9 +8,15 @@
       class="map-filters__filter-label hover--pointer flex flex-v-center"
       :class="{ 'map-filter--active': isActive}"
     >
-      <span :class="`map-filters__filter-key map-filters__filter-key--${config.id}`" />
-      <span class="map-filters__filter-title">
-        {{ name }}
+      <span
+        :class="[ !isEez ? `map-filters__filter-key map-filters__filter-key--${config.id}` : isActive ? 'map-filters__filter-key map-filters__filter-key--active':
```
stacytalbot, line 12: This can be re-written a bit dryer, rather than duplicating the classes that they both have you can have those by default and pull some of that logic out into a computed property so it is clearer to see what is going on in the template. (`:class="['map-filters__filter-key', eezClass, isActiveClass]"`)

app/javascript/components/map/filters/Dataset.vue
```diff
@@ -8,9 +8,15 @@
...
+        'map-filters__filter-key']"
+      />
+      <span
+        class="map-filters__filter-title"
+        :class="{ 'map-filters_filter-title_eezmap' : isActive }"
+      >
+        {{ isEez ? title : name }}
```
stacytalbot, line 19: Again you can pull the logic out the template into a computed property and just print out 1 variable here

PR #93, opened 2020-05-11
Small front end changes

stanleypliu:
https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/103


app/assets/stylesheets/resources/variables/_base.scss
```diff
@@ -143,7 +143,7 @@ $habitats: (coralreefs saltmarshes mangroves seagrasses coldcorals); // must mat
 $habitat_statuses: (present absent unknown);
 
 $habitat_colours: (
-  'coralreefs': $theme-pink,
+  'coralreefs': $color-coralreefs,
   'saltmarshes': $theme-purple,
```
stacytalbot, line 147: For consistency here I would change all these vars to be named after their sections and not the theme if thats the way you want to go rather than having some one way and some the other.

app/assets/stylesheets/resources/variables/_base.scss
```diff
@@ -29,14 +29,14 @@ $theme-purple: #332288;
 $theme-red: #E61354;
 $theme-yellow: #FCDA68;
 
-$theme-gradient-habitats: $green-light $theme-pink $theme-purple $theme-yellow $theme-blue $grey;
-
 $color-coralreefs: #CB47B5;
 $color-saltmarshes: $theme-purple;
 $color-mangroves: #D6A621;
 $color-seagrasses: $theme-blue;
 $color-coldcorals: $theme-red;
```
stacytalbot, line 36: These colours look like they need tidying up a bit - themes are probably from the older site. I would sort it out so that each hex code is attached to a var that describes its colour. then you can reuse that colour variable wherever you need to and its clear what colour it is.

app/helpers/countries_helper.rb
```diff
@@ -38,7 +38,8 @@ def total_value(dataset)
   end
 
   def total_units(dataset)
-    total_units = dataset[:id] == 'coldcorals' ? 'observations' : 'km<sup>2</sup>'.html_safe
+    # total_units = dataset[:id] == 'coldcorals' ? 'observations' : 'km<sup>2</sup>'.html_safe
+    'km<sup>2</sup>'.html_safe if total_units = dataset[:id] == 'coldcorals'
```
stacytalbot, line 42: Can commented out code be deleted? and is there a reason why this only applies to coldcorals and not the rest?

stanleypliu: Forgot to delete the commented out code, for some reason coldcorals was set to use observations instead of km2. I've now deleted the entire method and put `km<sup>2</sup>` in the actual view as there is now no need for a conditional.

app/assets/stylesheets/resources/variables/_base.scss
stanleypliu: I changed back to using $theme rather than naming via section, also tidied up _base.scss

PR #95, opened 2020-05-13
Fixed country map page not showing on refresh

stanleypliu:
Just deleted map.html.erb in countries/partials which was no longer being used and added id back into habitat_extent.html.erb to allow the country map to show


lucacug:
Looks the build errored, this happens without any reason sometimes, could you please rerun the build?

PR #98, opened 2020-05-27
Hide pages

stanleypliu:
About, Methodologies and the two Region pages have been hidden on the navigation bar.


app/views/global/_topbar.html.erb
```diff
@@ -3,8 +3,8 @@
     <%= link_to root_path, title: 'Visit the Ocean+ Habitats home page' do %>
       <%= image_tag 'logos/logo-ocean-plus-habitats.svg', class: 'logo--site', alt: 'The logo for Ocean+ Habitats' %>
     <% end %>
-    <%= link_to 'About', about_path, style: 'padding: 10rem' %>
-    <%= link_to 'Methodologies', methodologies_path %>
+    <%#= link_to 'About', about_path, style: 'padding: 10rem' %>
```
stacytalbot, line 6: @stanleypliu Not sure how that inline styling has appeared but when this is put back in can you change it to a class instead.

PR #99, opened 2020-05-27
Dials global page

stanleypliu:
Dials have been added to the home page with % protected for all habitats (values may need checking)


app/views/shared/_proportion-protected.html.erb
```diff
@@ -0,0 +1,46 @@
+<section id="<%= id %>" class="container section-padding--top">
+  <div>
+    <h2>
+      <%= t('shared.proportion_protected.title_html'); %>
+      <modal-trigger id="citation" :modal-content="<%= modal_content %>" class="icon--citation margin-space-left"></modal-trigger>
```
stacytalbot, line 5: Modal trigger doesn't need to be inside h2

Also see 'Change in habitat cover' section below where the title and icon are inline with each other.

PR #100, opened 2020-05-27
Refresh add sources

lucacug:
- Create Source model/migration and HABTM relation
- Create task to Import sources from csv
- Import geo entity stat sources 
- Import global static citations

~~:warning: May need some styling to the modals~~ split off into a separate PR


stanleypliu:
You may have addressed this already, but the modals still refer to the yml files which is governed by the `habitats_present_modal` method in `countries_helper.rb` which inserts placeholder text into the modals: 
image

PR #101, opened 2020-06-05
Modals styling

stanleypliu:
- Styled modals:
    - Increased horizontal padding to stop overlap of close button with text.
    - Shrunk radius of button and size of 'X'.
    - Made links live in modals and links open in new tab.
    - Shrunk font size. 
    - Added message for no citations.


lucacug:
@stanleypliu  looks good. Just a couple of citations icon are misaligned with the title (maps on the global page e.g.).  Otherwise happy to be merge.

PR #109, opened 2020-06-26
added sources header to every modal

stanleypliu:


app/javascript/components/modal/Modal.vue
```diff
@@ -14,7 +14,7 @@
           class="modal__close"
           @click="closeModal"
         />
-
+        <h3>Sources</h3>
```
WKocur, line 17: Can this go in the yml? You can then use the vue translation store to access the translation directly in the vue comp.

config/locales/home/habitat_cover/en.yml
```diff
@@ -10,27 +10,27 @@ en:
         coralreefs:
           title: Warm-water coral reefs
           text: ! '<strong>%{change_percentage}</strong> loss in warm-water coral cover<br>between <strong>%{baseline_year}</strong> and <strong>%{recent_year}</strong>'
-          citation: '<h3>Sources</h3><br>%{citation}'
+          citation: '<br>%{citation}'
```
WKocur, line 13: I think `<br>` can go. The citation could be wrapped in a `<p>` tag and actually doesn't need a yml entry anymore, so can all be done in the helper. Does that make sense?

app/javascript/components/modal/Modal.vue
stanleypliu: I could, but then my concern is that I would have to add this in for every YML file (as there is a separate YML file for each modal) so I would be repeating myself.

config/locales/home/habitat_cover/en.yml
stanleypliu: Yep, have done so, but - the helper method is acting on actual sources listed in the YML file for every other section, whereas the citations for the Habitat Cover section are actually being populated through the serialiser which is a special case - this is a result of new models being created specifically for these citations that are not country-specific and hence they are not present in the YML, but are rather being pulled directly from the database to populate the modals.

stanleypliu: It's not ideal to have two methods right now, but it's because these modals are specific to each habitat type, rather than covering the section as a whole.

app/javascript/components/modal/Modal.vue
WKocur: @stanleypliu You can put it in a global yaml file and then pull it into the component once, if that makes sense? 

See this component for an example: https://github.com/unepwcmc/ocean-plus-habitats/blob/47209bc0f95b571bdf351a12ea24919811b4ff3f/app/javascript/components/map/filters/MapDownloadButton.vue#L73-L86

app/helpers/application_helper.rb
```diff
@@ -108,11 +108,11 @@ def map_to_citations_string translations
       "<p>#{cit}</p>"
     end
 
-    citations.join
+    "<h3>Sources</h3>" + citations.join
   end
 
   def habitat_text(habitat)
-    habitat[:change_percentage].blank? ? I18n.t('home.habitat_cover.data_deficient') : habitat[:text].html_safe
+    habitat[:change_percentage] == '-' ? I18n.t('home.habitat_cover.data_deficient') : habitat[:text].html_safe
```
WKocur, line 115: @stanleypliu This is now going to show nothing when stats are missing - just checked with Osgur and Charlotte today.

app/serializers/serializers/habitat_cover_serializer.rb
```diff
@@ -12,59 +12,76 @@ def serialize
...
+
+    # Newline characters not properly recognised when converted to JSON
+    item[:modal_content] = "<h3>Sources</h3>" + arr.join("<br><br>")
+  end
+
   def reformat(habitat_global_change)
     if habitat_global_change.nil?
-      change_percentage = ''
-      citation = I18n.t('home.habitat_cover.no_citation')
+      change_percentage = I18n.t('home.habitat_cover.no_change_percentage')
```
WKocur, line 65: Will shown nothing when data is not present so will be an empty string (nil etc would work too)

config/locales/home/habitat_cover/en.yml
```diff
@@ -10,28 +10,19 @@ en:
...
           text: ! '<strong>%{change_percentage}</strong> loss in seagrass cover<br> between <strong>%{baseline_year}</strong> and <strong>%{recent_year}</strong>'
-          citation: '<h3>Sources</h3><br>%{citation}'
-          citation_url: '<a href="%{citation_url}">%{citation_url}</a>'
         coldcorals:
           title: Cold-water corals
           text: ! '<strong>%{change_percentage}</strong> loss in cold-water coral cover<br> between <strong>%{baseline_year}</strong> and <strong>%{recent_year}</strong>'
-          citation: '<h3>Sources</h3><br>%{citation}'
-          citation_url: '<a href="%{citation_url}">%{citation_url}</a>'
       data_deficient: Unfortunately, there is currently no data for this particular habitat.
+      no_change_percentage: '-'
```
WKocur, line 26: Given the previous comments, this shouldn't be needed anymore. Sorry, perhaps original it was meant to be a hyphen.

app/helpers/application_helper.rb
stanleypliu: Am I understanding correctly that we should just miss out that habitat altogether if stats are missing for it? As in, not show a separate card for it?

app/serializers/serializers/habitat_cover_serializer.rb
```diff
@@ -12,59 +12,76 @@ def serialize
...
+      else
+        link = citation[:citation_url]
+        citation_url = "<a target='_' class='modal__link' href='#{link}'>Source</a>"
+      end
+    
+      arr << citation_text + ' ' + citation_url
+    end
+
+    # Newline characters not properly recognised when converted to JSON
+    item[:modal_content] = "<h3>Sources</h3>" + arr.join("<br><br>")
```
WKocur, line 60: Is this how Charlotte/Osgur/Roger have asked for citation urls to be displayed? I'm not really that in the loop with this project.

WKocur, line 60: Can the citations each be a `<p>` element? Ideally a list, but that would be a bigger change. We try to avoid using <br> where possible and use margins for spacing.

stanleypliu: The above code relates to how the actual citations will be displayed (if there is more than one citation per habitat, to make it appear as if they are in a list), but in the light of the suggested changes, I'm going to rewrite my serializer and any associated code.

stanleypliu: Yep, I'm going to rewrite this now.

app/helpers/application_helper.rb
WKocur: No sorry, the card will show, but there won't be any text where the percentage would normally be. I've set up the templates so they should do this if the change percentage is blank

app/serializers/serializers/habitat_cover_serializer.rb
WKocur: Screenshot 2020-07-03 at 14 34 23
Oops! I tagged the wrong bit of code. I meant is this bit what it's meant to look like. Having `Source` as a link or `No URL found` at the end of the citation.

stanleypliu: Yep!

app/helpers/application_helper.rb
stanleypliu: I've changed the change_percentage to be nil now, and have updated the condition in the partial to reflect this.

stanleypliu:
The final thing I would say is something we may have missed: the habitat cover section has its own overarching modal at the top of the section next to the header which opens up the citations (which are present in the YML). Wondering if this is still necessary at all given that the habitats each have their own modal buttons? 
Screenshot 2020-07-03 at 15 34 00

WKocur:
> The final thing I would like to point out is: the habitat cover section has its own overarching modal at the top of the section next to the header which shows the citations that are present in the YML. Wondering if this is still necessary at all given that the habitats each have their own modal buttons?
> Screenshot 2020-07-03 at 15 34 00

Good question - I'm not sure. @Osgur ?

app/helpers/application_helper.rb
```diff
@@ -103,16 +107,17 @@ def habitats_present_modal
...
       "<p>#{cit}</p>"
     end
 
-    citations.join
+    "<h3>#{I18n.t('global.sources_modal_title')}</h3>" + citations.join
   end
 
   def habitat_text(habitat)
-    habitat[:change_percentage].blank? ? I18n.t('home.habitat_cover.data_deficient') : habitat[:text].html_safe
+    habitat[:change_percentage].nil? ? '' : habitat[:text].html_safe
```
WKocur, line 120: @stanleypliu This still shows the data_deficient text - if no data is present, it just won't show '-' next to the icon:
Screenshot 2020-07-06 at 08 34 01

stanleypliu: Ahhh, I deleted it. No worries I'll add that back in.

WKocur:
@stanleypliu Thanks for those fixes - I've had a look on the site and spotted that on a couple of the source links localhost has been added in before the external url e.g. this one:

Screenshot 2020-07-06 at 12 51 45

stanleypliu:
Have fixed now, the incorrect URLs were due to a mixture of unreadable characters and misplaced periods and my broken import method.

PR #123, opened 2021-05-24
Update 'Not known' to 'Data Deficient' in all places

stacytalbot:
https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/100

Update text in all places.


WKocur:
@stanleypliu I think this perhaps overlaps with something you were doing.

stanleypliu:
Thanks for pointing that out Will, just going to resolve the merge conflict like so for now then I'll sort it out properly on `refresh` directly.

PR #124, opened 2021-05-24
Feat/add missing country pages

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/69

Add all country pages (that have stats available) to the dropdown list. All country names are also in English, and should show the correct name in both the search dropdown as well as the actual page. 

Add a new pseudo-status, 'Present But Unknown' for habitats which are known to be present, but have no spatial data. They appear as 'Confirmed Presence' in the Habitats Present section, but 'Data Deficient' for all other sections. 

Also added the updated IUCN Redlist CSV as well as a CSV containing the updated list of countries, and the new habitat coverage statistics for each country and habitat. 

**Notes**

Make sure to get the new .env from Lastpass before you run any of the rake tasks below. 

You will need to run `rails db:migrate` then `rake import:refresh` locally to see the updated list of countries and also bring in the new habitat stats.


stanleypliu:
@WKocur Can be reviewed now

app/helpers/countries_helper.rb
```diff
@@ -1,29 +1,41 @@
...
   end
 
   def dataset_status(dataset)
-    dataset_status = @country.occurrences[dataset[:id]]
+    @country.occurrences[dataset[:id]]
   end
 
   def dataset_status_title(dataset)
-    dataset_status_title = I18n.t("shared.proportion_protected.title_#{dataset_status(dataset)}")
+    status = dataset_status(dataset) == 'present-but-unknown' ? 'data_deficient' : dataset_status(dataset)
```
WKocur, line 18: Only a small thing - if I'm calling a method twice, I usually assign it to a variable first:

```
status = dataset_status(dataset)
status_key = status == 'present-but-unknown' ? 'data_deficient' : status
```

Because this is such a simple method, so there aren't performance implications, up to you which you prefer :)

app/helpers/countries_helper.rb
```diff
@@ -1,29 +1,41 @@
...
   def icon_class(dataset)
-    icon_class = get_habitat_icon_class(dataset[:id], dataset_status(dataset))
+    status = dataset_status(dataset) == 'present-but-unknown' ? 'present' : dataset_status(dataset)
+
+    get_habitat_icon_class(dataset[:id], status)
   end
 
   def protected_percentage(dataset)
-    protected_percentage = @country.protection_stats[dataset[:id]] ? @country.protection_stats[dataset[:id]]['protected_percentage'].round(2) : 0
+    if @country.protection_stats[dataset[:id]]
```
WKocur, line 30: Likewise, here I would take `@country.protection_stats[dataset[:id]]` into a variable, but only a small thing. In this instance perhaps also a little easier to read.

app/helpers/countries_helper.rb
```diff
@@ -1,29 +1,41 @@
...
   end
 
   def protected_percentage(dataset)
-    protected_percentage = @country.protection_stats[dataset[:id]] ? @country.protection_stats[dataset[:id]]['protected_percentage'].round(2) : 0
+    if @country.protection_stats[dataset[:id]] 
+      protected_percentage = @country.protection_stats[dataset[:id]]['protected_percentage']
+
+      return 0 if protected_percentage.nil?
+
+      protected_percentage.round(2)
```
WKocur, line 35: Could do

```
protected_percentage.nil? ? 0 : protected_percentage.round(2)
```

Up to you

app/models/geo_entity.rb
```diff
@@ -15,9 +15,11 @@ class GeoEntity < ApplicationRecord
 
   has_many :country_citations, foreign_key: 'country_id'
 
-  scope :countries, -> { where.not(iso3: nil) }
+  scope :countries, -> { where.not(iso3: nil || 'GBL') }
```
WKocur, line 18: Is this Global?

app/models/geo_entity.rb
```diff
@@ -32,19 +34,58 @@ def count_species
...
+
     hash = {}
 
-    _occurrences.map do |occurrence|
-      _occurrence = occurrence['occurrence']
+    occurrences.map do |occurrence|
       next if hash[occurrence['name']] == 'present'
-      hash[occurrence['name']] = _occurrence unless _occurrence == 'unknown'
+
+      # unless occurrence['occurrence'] == 'unknown'
```
WKocur, line 52: Can this go now?

app/models/geo_entity.rb
```diff
@@ -32,19 +34,58 @@ def count_species
...
 
+  def fetch_needed_occurrence_attrs
+    countries_geo_entity_stats.joins(:habitat).select(
+      'habitats.name, 
+      occurrence, 
+      protected_value, 
+      total_value, 
+      geo_entity_stats.protected_percentage'
+    )
+  end
```
WKocur, line 67: Can this go below the `determine` methods so it reads high level functions to low level?

app/models/geo_entity.rb
```diff
@@ -32,19 +34,58 @@ def count_species
...
+
+  def determine_original_occurrences
+    occurrences = fetch_needed_occurrence_attrs
+
+    occurrences.map(&:attributes).map { |attrs| attrs.slice('name', 'occurrence') }
```
WKocur, line 72: Does this do the same as 

```
occurrences.as_json(only: [:name, :occurrence])
```
?

(as_json is a weird naming... because I think it returns a ruby hash 🤷 )

WKocur: You might get string keys instead of symbols... not sure

app/serializers/serializers/map_datasets_serializer.rb
```diff
@@ -113,7 +113,7 @@ def serialize
     DATASETS.map do |ds|
       dataset = ds.dup
       dataset[:name] = get_habitat_from_id(ds[:id])[:title]
-      dataset[:disabled] = habitat_presence_status(ds) == 'unknown' || habitat_presence_status(ds) == 'absent'  ? true : false
+      dataset[:disabled] = GeoEntity::NEGATIVE_OCCURRENCE_STATUSES.include?(habitat_presence_status(ds)) ? true : false
```
WKocur, line 116: Do you need `? true : false` at the end?

app/views/countries/partials/_habitats-present.html.erb
```diff
@@ -2,17 +2,25 @@
...
-  <p><%= t('countries.shared.habitats_present.intro_text', country_name: t("countries.#{@yml_key}.title")); %></p>
+  <p><%= t('countries.shared.habitats_present.intro_text', country_name: @name); %></p>
 
   <div class="cards--habitats-present">
     <% habitats_present.each do |habitat| %>
+      <!-- Present but data deficient habitats are essentially considered as 'Present' -->
+      <% if habitat[:status] = 'present-but-unknown'  %>
+        <% habitat[:status] = 'present' %>
+        <% habitat[:status_title] = 'Confirmed Presence' %>
+      <% end %>
```
WKocur, line 13: Can this be incorporated into a helper or similar? Perhaps in the HabitatsPresentSerializer (if that's where this is coming from)

app/views/shared/_proportion-protected-chart.html.erb
```diff
@@ -16,7 +16,7 @@
           <p class="card__habitat-title"><%= dataset[:name] %></p>
           <p class="card__habitat-status">
             <% if @habitats_present %>
-              <% if (dataset_status(dataset) == 'unknown' || dataset_status(dataset) == 'absent') %>
+              <% if GeoEntity::NEGATIVE_OCCURRENCE_STATUSES.include?(dataset_status(dataset)) %>
```
WKocur, line 19: Could this be turned into a helper that takes dataset as the argument? Might be able to use it elsewhere too. Don't worry if that requires a lot of refactoring!

GeoEntity::NEGATIVE_OCCURRENCE_STATUSES.include?(dataset_status(dataset))

config/locales/shared/proportion_protected/en.yml
```diff
@@ -7,5 +7,6 @@ en:
       title_present: Confirmed Presence
       title_absent: Confirmed Absence
       title_unknown: Not Known
+      title_data_deficient: Data Deficient
```
WKocur, line 10: Is the duplication here necessary? Is it needed in countries/shared, if it's in shared?

lib/tasks/import_bounding_boxes.rake
```diff
@@ -20,4 +20,31 @@ namespace :import do
...
     end
   end
+
+  def import_bounding_boxes_from_esri
+    filename = "lib/data/habitat_presence/habitat_presence_country.csv"
+
+    CSV.foreach(filename, headers: true) do |row|
+      actual_name = row['COUNTRY']
+      iso3 = row['ISO3']
+      geo_entity = GeoEntity.find_by(actual_name: actual_name, iso3: iso3)
```
WKocur, line 30: Would iso3 on its own be sufficient?

lib/tasks/import_bounding_boxes.rake
```diff
@@ -20,4 +20,31 @@ namespace :import do
       end
     end
   end
+
+  def import_bounding_boxes_from_esri
+    filename = "lib/data/habitat_presence/habitat_presence_country.csv"
```
WKocur, line 25: I've always seen Rails.root used for pathnames, not sure if it's needed. Will this work on staging and production?

lib/tasks/import_countries.rake
```diff
@@ -1,26 +1,29 @@
...
 
 namespace :import do
-  desc "import countries data into database"
-  task :countries => [:environment] do
+  desc 'import countries data into database'
+  task countries: [:environment] do
+    # Strip out accents, commas, apostrophes, brackets as they mess with the URL
+    def sanitize(name) 
+      I18n.transliterate(name).downcase.gsub(/[,'()]/, '').titleize
+    end
```
WKocur, line 9: Does this mean we don't need the pathname gsub going on somewhere else in this PR?

lib/tasks/import_prebakedstats.rake
```diff
@@ -20,61 +20,65 @@ namespace :import do
   def import_new_csv_file(habitat, csv_file)
     filename = "#{Rails.root}/lib/data/#{csv_file}"
     CSV.foreach(filename, headers: true, encoding: "utf-8") do |row|
+      name = row['ISO3'] || row['region']
+      geo_entity = fetch_geo_entity(name)
+      next unless geo_entity
+
       if row.select {|k,v| /baseline/i =~ k}.any?
```
WKocur, line 27: Could this condition be in a named method? It isn't clear to me what it's doing/why. Excluding all rows without an entry for a baseline column?

lib/tasks/import_prebakedstats.rake
```diff
@@ -83,7 +87,9 @@ namespace :import do
       geo_entity = GeoEntity.find_by(name: "Disputed")
     else
       geo_entity = GeoEntity.find_by(iso3: name) || GeoEntity.find_by(name: name)
+      return unless geo_entity
```
WKocur, line 90: What's the difference between returning nothing here and waiting and then returning nil in the line below?

WKocur (changes requested):
Looks like it's working really well and the code makes sense and is easy to read 🎉 

Ran through and have added a few comments - most are small refactors and some are just suggestions for alternatives if you happen to prefer them. A couple of general things from looking at the site and testing a few countries at random:

- [ ] Aland islands doesn't load. TO BE DEALT WITH ON ANOTHER PR/TICKET
- [x] Add svgs with iso names
- [x] It's losing accents on some characters in the country names: Sao Tome And Principe. Is that intended? TO BE DEALT WITH ON ANOTHER PR/TICKET
- [x] Proportion of habitat protected for http://localhost:3000/sao-tome-and-principe has a lot of 0%. Is that correct?

P.S. Homepage doesn't load but I'm assuming that's not part of this PR.
P.P.S I haven't checked the csv data to see if the figures showing on the site are correct.

stanleypliu:
Good point about the Åland Islands, they along with 5 other countries/territories currently have no coastal stats (not habitat stats) associated with them, but hopefully CLS should be able to provide them. So they are broken for the time being.

You're right that some countries are losing accents when they shouldn't be, but it's affecting certain diacritics like ã for some reason. Thanks for pointing that out, looking into it 👍 

Yes, Sao Tome and Principe has a lot of 0% which is technically correct due to the rounding, it does actually have those habitats but they are so small in area the app rounds them. The only one which seems off is the Cold Water Coral coverage, which should read '0% of 1 km2', I can see why that is happening. 

Homepage not loading is also part of this PR due to the changes I've made so I'm looking into that too

README.md
```diff
@@ -25,3 +27,6 @@ which is updated every month.
 Habitats data changes less often, so the coverage calculations are stored in the `Habitat` object.
 When the data changes, a new calculation process should be performed using the `calculate_global_coverage` method in
 the `Habitat` model.
+
+Old coverage data is stored within `lib/data/countries` for the various habitats. The new coverage data can be found within
```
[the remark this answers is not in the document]
stanleypliu: I'm not sure - there is a significant amount of historical data here from when the app was first built which wasn't removed, so I'm on the side of keeping it and just making a note that it exists. If for whatever reason we need to retrieve it, it'll be easier than diving into the history IMO

app/controllers/countries_controller.rb
```diff
@@ -3,9 +3,11 @@ class CountriesController < ApplicationController
 
   def show
     @country = GeoEntity.find_by(name: country_name_from_param(params[:name]))
-    @yml_key = @country.name.downcase.gsub(/ /, '_').gsub('%27', "")
 
-    country_yml = I18n.t("countries.#{@yml_key}")
+    # TODO - work out how to integrate i18n with this for country names - we may want to
+    # have a list of country names in each language which is then dynamically fetched from
+    # a CSV/yml depending on the language selected
+    @name = @country.actual_name
```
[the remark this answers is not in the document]
stanleypliu: I think only English for now, but good to know about Globalise. IIRC they are planning to add German and French/Spanish to the site as well. 

I was looking into a gem for country name translations but it only supports fetching via ISO2, not ISO3 unfortunately.

app/helpers/countries_helper.rb
```diff
@@ -1,29 +1,41 @@
 module CountriesHelper
-  def if_country_get_flag yml_key
-    country = I18n.exists?("countries.#{yml_key}.image_flag")
-    flag = FileTest.exist?("app/assets/images/#{t("countries.#{yml_key}.image_flag")}")
+  def if_country_get_flag(country_name)
+    name = country_name.downcase.gsub(' ', '-')
```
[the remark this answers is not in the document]
stanleypliu: Yeah good point. If we go down that route then, are you going to be using the ISO3 codes for the flag filenames then? If so, I might change this method now

app/controllers/countries_controller.rb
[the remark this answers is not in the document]
WKocur: I think we have an iso2 to iso3, if wanted, although I suspect the list would have to be of UN approved names, so would need to be custom... Not sure!

app/helpers/countries_helper.rb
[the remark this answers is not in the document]
WKocur: That would make sense to me I think yeah 👍  Cheers!

lib/tasks/import_bounding_boxes.rake
stanleypliu: Yes, it should work on staging and production (and it has done actually since this functionality to import the bounding boxes is already on staging).

lib/tasks/import_countries.rake
stanleypliu: I'll have to take a look to see if any refactoring can be done in the helpers

lib/tasks/import_prebakedstats.rake
stanleypliu: Yeah you might be right in that, because if the second half of the OR condition returns `nil`, then `geo_entity` will be `nil` anyway.

app/helpers/application_helper.rb
```diff
@@ -27,24 +27,35 @@ def social_image_alt
...
+      nav_item(country.actual_name) 
+    end
+  end
+
+  def nav_item(name)
+    path_name = name
+    geo_entity = GeoEntity.find_by_actual_name(name)
+    geo_entity_id = geo_entity&.id
+    
+    return {} unless geo_entity_id
```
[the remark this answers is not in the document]
stanleypliu: There is the `reject` method for this sort of thing. I feel as though that would still require me to subsequently add a method to the end of `list_of_countries` to do the filtering, so it wouldn't really add much to be honest.

app/models/geo_entity.rb
stanleypliu: Yep - it's created as part of the Red List import task. Still unclear about what to do regarding that - it definitely needs fixing as one of the next tickets and I may even remove this change as part of that.

app/models/geo_entity.rb
stanleypliu: Done, placed within private methods

app/models/geo_entity.rb
stanleypliu: After looking at my methods, I realise that the `determine_original_occurrences` probably isn't even needed anymore (and it's definitely not being called) with this new approach. I'm just going to remove it altogether. 
As for applying the `as_json` method, you're right that  `map(&:attributes)` and then `slice` is equivalent to it and it is shorter, but it's misleading to read as we're not really outputting a JSON, we're outputting a stringified hash.

app/serializers/serializers/map_datasets_serializer.rb
stanleypliu: Nice catch

app/views/shared/_proportion-protected-chart.html.erb
stanleypliu: Yep I'm going to place this in the Countries Helper for now. To be honest, I think we need a presenter at this rate to deal with the amount of logic in the view files.

lib/tasks/import_countries.rake
stanleypliu: No, we still need that method, as that basically kebab-cases the country name. It's true that we don't need to account for apostrophes any more though.

lib/tasks/import_prebakedstats.rake
stanleypliu: I've added a comment to explain

app/helpers/countries_helper.rb
```diff
@@ -1,9 +1,10 @@
 module CountriesHelper
-  def if_country_get_flag(country_name)
-    name = country_name.downcase.gsub(' ', '-')
-    flag = FileTest.exist?("app/assets/images/flags/#{name}.svg")
+  def if_country_get_flag(country_iso3)
```
WKocur, line 2: I think you forgot to pass the iso_3 to the method in the view! Otherwise looks good 👍

PR #125, opened 2021-05-25
Refresh hero

WKocur:


app/helpers/application_helper.rb
```diff
@@ -135,7 +135,16 @@ def habitat_protection(name)
...
   end
+
+  def get_nav_primary
+    [
+      { 
+        title: 'About',
+        url: about_path,
+      }
+    ].to_json
+  end
```
WKocur, line 149: Should these be in translations?

PR #131, opened 2021-05-27
Section structure

WKocur:
https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/129


app/assets/stylesheets/styles/components/cards/_cards-habitats-present.scss
```diff
@@ -1,7 +1,7 @@
 &--habitats-present {
   @include flex;
   @include flex-h-between;
-  margin-bottom: $page-padding-bottom-small;
```
stacytalbot, line 4: Merging - the only thing i'll say here is that these padding variables were created to keep spacing between components consistent.

PR #140, opened 2021-05-28
chore: update coastline coverage statistics

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/149

Updated coastline coverage CSV

**Notes**

Run `rake import:coastalstats`


lib/data/coastline_coverage/coastline_coverage.csv
```diff
@@ -1,613 +1,640 @@
...
+"VIR",364.342620623377,0,0,0,0,0,363.919780968217,50
+"VNM",15006.3356871535,2029.68656913059,582.595436980978,10.5560463527137,756.061874753668,0,10712.211795523,50
+"VUT",3049.69943579994,0,0,0,871.461883557195,0,2172.37060455036,50
+"WLF",302.31477979448,0,0,0,0,0,301.733369469651,50
+"WSM",449.302732232574,0,0,0,0,0,448.766639055566,50
+"xHT",309.936958415584,0,0,0,71.9469023663863,0,237.560725624442,50
+"xKI",1263.82808857718,0,0,285.107675391024,0,120.165150203047,104.191896209455,50
+"xPI",155.148819782731,0,0,0,133.450474485418,0,21.6099649032627,50
+"xSI",18.8032133103445,0,0,0,17.2318494174414,0,1.54370852231876,50
+"xSK",19.0163705162395,0,0,0,19.0107420287208,0,0,50
```
WKocur, line 638: I'm guessing these are just ignored, seeing as they are not real ISO3 codes (as far as I'm aware)? And I guess the second question is, should they be ignored?

WKocur:
Should this fix the error when trying to load Aland Islands?

WKocur:
Oh wait - ignore. Just seen the other PR

lib/data/coastline_coverage/coastline_coverage.csv
stanleypliu: I can check this with CLS - this could potentially be some kind of non-ISO3 code (but still ISO) - best to keep this in case we need it?

WKocur: That would be great if you don't mind. I'm happy to merge this anyway as I don't think it's doing any harm, may just mean they need to supply a new csv if this data is meant to match up with a GeoEntity that doesn't currently exist/has a different code.

PR #142, opened 2021-05-28
Fix/broken country pages

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/150

What I thought was a lack of coastline coverage statistics that were breaking certain country pages, was in fact a lack of Redlist data and other assorted data like occurrences for particular habitats. So this PR fixes this by: 
-  adding a method to the `import:prebaked stats` Rake task that trawls through the `habitat_presence_country` CSV and updates any statistic without an occurrence using it.
- hiding both the Redlist table and the example selection component if no Redlist data is present. 

**Notes**

Run `rake import:prebakedstats` or if that doesn't work, `rake import:refresh`
Go to pages for Aland Islands, Hong Kong, Isle of Man, even Antarctica (technically not a country, but that's probably another issue) and they should be working. 

‼️ If the code works then don't merge this straight away please - I need to deploy this branch to staging for CLS to review so they can suggest any changes.


app/views/countries/partials/_example-species.html.erb
```diff
@@ -24,10 +26,15 @@
     <p class="text--smallprint max-width-container-small">
       <%= t('countries.shared.example_species.smallprint').html_safe %>
     </p>
+  <% else %>
+    <div class="h2-wrapper flex flex-v-center">
+      <h3>No species data is present for this particular country/territory.</h3>
```
WKocur, line 31: Can this be moved to a yml file as well?

lib/tasks/import_prebakedstats.rake
```diff
@@ -15,6 +15,10 @@ namespace :import do
         import_new_csv_file(habitat.name, csv_file)
       end
     end
+
+    # Import occurrences for any countries that are missing (disregarding regions for the moment)
+    # using habitat_country_presence.csv
```
WKocur, line 20: I'm not sure I understand why there are missing occurrences/why running whichever importer usually imports these values wouldn't fix this.

WKocur (commented):
Aland islands loads nicely! Of course, how it handles this missing data will need to be checked by Roger/CLS if it hasn't already, but good to have the pages loading!

lib/tasks/import_prebakedstats.rake
stanleypliu: This was a little tricky to solve, but I pinpointed it to the fact that there is missing species data for certain countries in the Redlist CSV (as in there are no rows in that CSV with the ISO3s of those countries, hence no species data), but which have habitat coverage data present (which is provided by `habitat_country_coverage` which has the full list of countries/territories). 

As we only get a portion of the full statistics in each CSV/set of CSVs, I decided to make this part of the 'prebaked' stats task since it directly deals with creating and updating GeoEntityStats, which are the crucial model that the statistics rely on to populate country pages. In this case, after the main import is run with the CSVs in the `habitat_coverage_protection` folder, I then use another CSV, the `habitat_presence_country` CSV, which contains all of the countries, plus their occurrences. *But*, there is no data on the coverage and protected area for each habitat in this CSV, it only exists in the previous set of CSVs. Hence I have to use both sources to fully populate the statistics. Basically, I fetch any GeoEntityStats with a value of `nil` for the `occurrence` attribute to trace the parent country, grab the ISO3 of the country and then use the `occurrence` values in `habitat_presence_country` to update each habitat in turn. 

Hope that makes sense

stanleypliu: To be honest, it's better that all of this information is combined into one CSV, but that isn't the case

WKocur: I think I follow! Just to check - if you run db:seed on a refresh db, will this import everything?

stanleypliu: Yes it should

WKocur:
@stanleypliu Happy for you to merge this when ready, if it's all fine after being checked on staging 👍

stanleypliu:
Still going to keep the conditional checks in place for lack of data, just in case

PR #145, opened 2021-05-31
Test suite

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/40

Unit tests in place for Species, GeoEntity, GeoEntityStat.
Request specs in place for Country and Site controllers. 

I've also added the Faker gem and have removed a few redundant methods that I spotted in the models.
Also present is a Rake task to allow linting of future FactoryBot factories.

Updated Travis file to use RSpec instead of Minitest. 

Add Codecov gem to get reports and badge.

**Notes**

Run `bundle install` and then `bundle exec rspec`. All tests should be passing. 

Also note that the ENV var, `ESRI_SERVER_EEZ_URL` has been changed here and also on Lastpass.


stanleypliu:
All of the tests pass locally for me but there are two tests that are failing when run via Travis. Investigating 🔍

spec/models/geo_entity_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+      countries.each do |country|
+        FactoryBot.create(:geo_entity_stat, habitat: habitat, geo_entity: country)
+      end
+    end
+  end
+
+  let(:geo_entity) { GeoEntity.last }
+
+  describe '#count_species' do
+    it 'calls the relevant method on the Species model' do
```
WKocur, line 17: Took me a while to work out what was being tested here. Would it be against convention for the description to be 'calls count_species on the Species model when calling count_species on itself'. 

It sounds a bit over the top, but I think I was confused because of the two methods being named the same and I missed for a while that they were defined on different models... Not a big deal, just might save time for people looking at it in the future. Ignore if you think that makes it more confusing!

WKocur (commented):
I've only given this a very basic look over. Everything seems to work fine, but it might be worth getting someone more familiar with rspec and factorybot to look over this in the future. For now, we're a bit time pressed - but looks good to me!

Should these now be passing on Travis?

spec/models/geo_entity_spec.rb
stanleypliu: The way this method works is a bit convoluted in that it's calling a method that exists on the Species model, which shouldn't really be happening as ideally that could all be handled within the Species model itself. I might look at removing this spec (and refactor that particular method along with any method calls to it) in the future.

spec/models/habitat_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      expect(habitat.calculate_global_cover_change).to eq({ baseline_year: 2010, change_km: 0.4e0, change_percentage: 0.8e2, original_total: 0.5e0 })
+    end
+
+    it 'returns the correct statistics' do
+      expect(habitat.global_stats).to eq({ protected_habitat_cover: 0.38e2, total_habitat_cover: 0.1e3 })
+    end
+
+    it 'calculates the global protection' do
+      expect(habitat.calculate_global_protection).to eq({ 'protected_percentage' => 38.0, 'total_value' => 100.0 })
+    end
```
WKocur, line 49: Do you think the expected values should be calculated from test data? I realise that then there's untested logic in the tests... I can see someone changing the test data and then the tests will break. Maybe that's fine though. I'm undecided on this! Do you know what is standard/best practice?

WKocur:
All passing for me!

Is this test meant to be skipped:

```
  1) Country pages Access a country page
     # Temporarily skipped with xdescribe
     # ./spec/system/country_pages_spec.rb:11
```

spec/models/habitat_spec.rb
stanleypliu: Yeah this might be flaky test logic - I can have another look at this.

stanleypliu:
I've skipped it deliberately because it took a while to configure the set up properly to allow it to render Vue components properly. I think it's fine now so will be creating a couple of system specs today.

spec/models/habitat_spec.rb
stanleypliu: I had another look at this and have now replaced static values in the Habitat spec with the record values from the fixtures.

spec/spec_helper.rb
```diff
@@ -17,7 +17,12 @@
 # See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
 RSpec.configure do |config|
   config.before(:each, type: :system) do
-    driven_by :selenium, using: :chrome, options: { args: ["headless", "disable-gpu", "no-sandbox", "disable-dev-shm-usage"] }
+    driven_by :selenium, using: :headless_chrome, options: { 
+      args: ["headless", "disable-gpu", "no-sandbox", "disable-dev-shm-usage"],
+      prefs: {
+        'download.default_directory' => "#{Rails.root}/tmp/capybara_downloads"
+      }
```
WKocur, line 24: What's the difference between spec_helper and rails_helper? Are they for different types of tests? Seems like selenium is being configured twice, but perhaps for different purposes?

spec/system/country_page_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  scenario 'Finding the total length of the coastline' do
+    chart_title = find('.habitat-representation__chart-title')
+
+    unrounded_coastline = GeoEntity.last.coastal_stat.total_coast_length
+    coastline_length = ActiveSupport::NumberHelper.number_to_delimited(unrounded_coastline.round)
+
+    expect(chart_title.text).to eq("Percentage of coastline (#{coastline_length} km) covered by habitats")
+  end
```
WKocur: I just want to check this test is testing what it should. Is it an integration test? This test would pass if total_coast_length was calculating the wrong stat (which I'm sure you know). So what it's testing is that the correct methods and formatting are being used to populate the chart_title. Is that all it should be testing? Not sure what convention is here, so just checking :)

spec/system/country_page_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  scenario 'Looking at example of species section' do
+    expect(page).to have_content('Total number of species: 20')
+
+    species_common_names = GeoEntity.last.species.take(3).pluck(:common_name)
+
+    expect(page).to have_content(species_common_names[0])
+    expect(page).to have_content(species_common_names[1])
+    expect(page).to have_content(species_common_names[2])
```
WKocur, line 36: Perhaps these should target the elements where you expect that content to be - or at least limit it to the component container. That would be more robust, in case in the future, species names appeared elsewhere on the page. Just a small thing.

spec/system/country_page_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  scenario 'Opening the main national data sources modal' do
+    find('.downloads__li > #modal-trigger-citation').click
+
+    expect(page).to have_css('.modal__dialog')
```
WKocur, line 42: Is there a way to easily check that the modal__dialog is visible? E.g. if it were changed to use v-show for some reason, would this then always pass? I'd be tempted to check that it isn't visible before and then is visible afterwards, to be sure that a pass corresponds to it working properly.

Also, is it worth checking for the 'Sources'  title (or even better one of the sources), to check that it isn't some other modal that has accidentally been opened (which could happen if the js was broken as the modals are shared).

spec/system/country_page_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  scenario 'Downloading set of national statistics as a zip' do 
+    download_link = find('a', text: 'Download statistics')
+
+    root_url = page.current_url.delete_suffix(country_path)
+    expect(download_link[:href]).to eq("#{root_url}downloads/national/#{GeoEntity.last.iso3}.zip")
```
WKocur, line 49: Is it not easy to test that it downloads the file when clicked?

spec/system/home_page_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  scenario 'Accessing the country dropdown' do 
+    visit '/'
+
+    find_all('#v-select-search-country').first.click
+
+    expect(page).to have_all_of_selectors(
+      "#option-country-#{GeoEntity.first.id}", 
+      "#option-country-#{GeoEntity.second.id}",
+      "#option-country-#{GeoEntity.third.id}"
+    )
```
WKocur, line 15: Would it be useful to also check the total number of options?

spec/system/home_page_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  scenario 'Downloading set of global statistics as a zip' do 
+    visit '/'
+
+    download_link = find('a', text: 'Download statistics')
+
+    expect(download_link[:href]).to eq("#{page.current_url}downloads/global_statistics.zip")
```
WKocur, line 23: Same comment as on other download test in country_page_spec

spec/system/home_page_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  # Not actually going to try to download the shapefiles, they are massive!
+  scenario 'Opening the spatial data downloads modal' do
+    visit '/'
+
+    find('#modal-trigger-spatial-downloads').click
```
WKocur, line 30: Same comment as before

spec/system/home_page_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    expect(page).to have_css('.modal__dialog')
+
+    # Can be flaky because of the dependence on the very specific `for` attribute
+    page.find('label[for="download-radio-buttons-56-option-saltmarshes"] > .radio-button__radio').click
```
WKocur, line 35: Is this because the 56 presumably isn't always 56? Not sure how Vue's _uid is calculated... Can you match based on regex, so you don't care about the number?

spec/system/home_page_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+    visit '/'
+
+    expect(page).to have_css('#eez_map')
+
+    expect(page).to have_css('.map-filters__filter.map-filters__filter--eez')
+
+    page.find('.map-filters__filter-key.map-filters__filter-key--eez-coldcorals').click
+
+    expect(page).to have_css(
+      '.map-filters__filter.active.map-filters__filter--eez > label > .map-filters__filter-key.map-filters__filter-key--eez-coldcorals'
```
WKocur, line 61: I guess this framework doesn't support testing whether js methods are called, so it's hard to get at whether the layer is actually added? Potentially mapbox emits an event when adding a layer... not sure if you could listen to that via this framework...

WKocur (commented):
Some comments - my experience with tests was in a different framework and was testing frontend only (albeit a frontend that had it's own MVC thing going on), so I'm not really sure on how these should work, so if I'm saying something silly, just tell me! Good to see some comprehensive tests :D

spec/spec_helper.rb
stanleypliu: They're both files generated by RSpec, one is deliberately meant to be more customisable (the RailsHelper) and it holds more configuration as a result, also because for standard Rails classes like controller and unit tests. SpecHelper is a more generic helper that does not depend on Rails and can therefore be used for testing files in `lib` for example, which don't have access to the Rails environment unless specified. It has that extra config for Selenium because it's explicitly setting Selenium as the driver for system specs, whereas RailsHelper contains the general driver configuration.  

https://stackoverflow.com/questions/24145329/how-is-spec-rails-helper-rb-different-from-spec-spec-helper-rb-do-i-need-it

spec/system/country_page_spec.rb
stanleypliu: It's just checking that it actually shows the statistic in the chart title on the page (regardless of any styling used or whether the statistic is wrong). But it will break if the class is changed. Unfortunately that's a problem with these system specs.

spec/system/country_page_spec.rb
stanleypliu: It isn't unfortunately, I decided it wasn't worth downloading a file each time this method is called, because then the file would need to be deleted to avoid it persisting. Unfortunately, the download folder doesn't seem to be set properly so it goes straight into the system's Downloads folder. This method is easier - it checks for a link with the correct name and href.

spec/system/home_page_spec.rb
stanleypliu: Correct - it's because the method that's generating the number is using _uid. I can have a look at matching with Regex though

spec/system/home_page_spec.rb
stanleypliu: Capybara unfortunately doesn't offer the option to detect these events, and even then, this method I felt was easier than listening to events, because I knew that it would always add an `active` class to the selected option.

spec/system/home_page_spec.rb
stanleypliu: https://stackoverflow.com/questions/14635474/how-to-use-regex-in-a-capybara-finder this link is pretty useful re: CSS selectors

spec/system/home_page_spec.rb
```diff
@@ -37,8 +37,7 @@
     
     expect(page).to have_css('.radio-group.download-radio-buttons__radio-group > li.radio-group__item', count: 5)
 
-    # Can be flaky because of the dependence on the very specific `for` attribute
-    page.find('label[for="download-radio-buttons-56-option-saltmarshes"] > .radio-button__radio').click
+    page.find('label[for*="saltmarshes"] > .radio-button__radio').click
```
WKocur, line 40: Just a teeny one: Would `downloads_dialog.find('label[for*=...` work here (can you do that?)? I guess it's feasible that in the future other inputs for the habitats could be added to the page, which could mess with this test.

PR #148, opened 2021-06-02
update footer re downloads and legal link

WKocur:
https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/117


nym1k (changes requested):
The download buttons are not centred as per the design. Also, in the design the 'View list' buttons do not have an icon (Unless this has changed in a new design)

WKocur:
@nym1k this changes in #154 so can probably be left until that one. As discussed in standup, I think the text used here will change.

PR #159, opened 2021-06-04
feat: adding search bar to home hero

nym1k:
Add 'country or territory' search bar to home hero

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -170,13 +170,6 @@ module ApplicationHelper
     }.to_json
   end
 
-  def get_countries_hero_search_config
-    {
-      id: 'country',
-      placeholder: 'Search for Country or Territory'
-    }.to_json
-  end
-
   def get_nav_primary
     [
       {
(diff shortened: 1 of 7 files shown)


app/helpers/application_helper.rb
```diff
@@ -170,9 +170,16 @@ def get_countries_search_config
     }.to_json
   end
 
+  def get_countries_hero_search_config
+    {
+      id: 'country',
+      placeholder: 'Search for Country or Territory'
```
stanleypliu, line 176: Do you know if this is a standard input element or does it use some custom logic which requires it to be set like this? If it is a standard input I think you can just set the `placeholder` attribute directly on the Vue component

nym1k: Removed, and hard coded with translation

app/views/global/_header-home.html.erb
```diff
@@ -8,7 +8,10 @@
   <div class="header__search">
     <v-select-searchable
       class="topbar__dropdown--centered v-select--large"
-      :config="<%= get_countries_hero_search_config %>"
+      :config="<%= {
+        id: 'country',
+        placeholder: t('home.general.search_placholder')
```
stanleypliu, line 13: Small typo here, but otherwise works nicely, I will merge when it's done

PR #164, opened 2021-06-07
Chore/update ocean plus static citations

stanleypliu:
No ticket, but this is from the updated OceanPlus_static_citations csv that was provided (and is present in this PR). 

I've basically gone and updated, using that CSV as a reference, the sources present in each modal across the site. One notable change is that the EEZ map now has a distinct set of references from the Proportion Protected section on the home page. 

**Notes**

Potentially still need to change the Sources for the 'Examples of species and their status for habitat' section on the country pages. This is in reference to the last two rows of the CSV, which apply specifically to that section.


WKocur:
@stanleypliu Looks good to me - only thing is that we might want to find and replace 'doi' with 'DOI' - to be confirmed in standup.

PR #156, opened 2021-06-03
Feat/national csv downloads

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/120

This creates within the `public` folder, a separate CSV for mangroves and a CSV for all other habitats for that country. 

**Notes**

Run `rake generate:national_csvs` to generate the CSVs for each country.

PR #165, opened 2021-06-07
Ticket #156 - 404 Page

nym1k:
Custom 404,500 error pages
New routes for errors with controller and views


app/controllers/countries_controller.rb
```diff
@@ -4,6 +4,8 @@ class CountriesController < ApplicationController
   def show
     @country = GeoEntity.find_by(name: country_name_from_param(params[:name]))
 
+    return redirect_to(action: 'unprocessable_entity', controller: 'errors') unless @country
```
WKocur, line 7: I would probably redirect to not_found here. What do you think?

nym1k: Agreed (think I missed this after testing)

WKocur:
I think you need this `config.exceptions_app = self.routes` in `application.rb`. You can test what it will do on prod by setting `config.consider_all_requests_local` to false in `environments/development.rb`

WKocur:
Have you tested an incorrect url that doesn't go to the country controller? That is what didn't work for me.

PR #166, opened 2021-06-07
Ticket #154 - Update footer copyright info

nym1k:
Updating the footer copyright text in the global locale file

### DIFF
--- a/config/locales/global/en.yml
+++ b/config/locales/global/en.yml
@@ -1,6 +1,6 @@
 en:
   global:
-    page_title: 
+    page_title:
       index: Ocean+ Habitats
       about: About
       legal: Legal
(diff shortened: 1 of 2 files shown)


config/locales/global/en.yml
```diff
@@ -35,14 +35,14 @@ en:
         icon: icons/cold-coral.svg
     nav:
       burger_text: Explore a country/region
-    
+
     footer:
       about: About Ocean+ Habitats
       legal: Legal
-      copyright: Global Ocean Habitat Profile from the Ocean+ &copy; copyright
+      copyright: '&copy; 2021 UN ENVIRONMENT PROGRAMME (UNEP)'
```
WKocur, line 42: Can we add the 2021 with Date.today.year in the rails view using the i18n interpolation? Currently you have 2021 twice!

PR #167, opened 2021-06-07
Ticket #116 - Citation Modal Tooltip

nym1k:
Add tooltip to modal trigger
Sources modal title now sits under Sources modal group in global locales
View sources tooltip text added to locales


WKocur:
Works nicely - no changes needed :)

PR #172, opened 2021-06-08
fallback to solid fill if layer image not found

WKocur:
Shows solid fill if image isn't found.

### DIFF
--- a/app/javascript/components/map/mixins/mixin-add-layers.js
+++ b/app/javascript/components/map/mixins/mixin-add-layers.js
@@ -5,7 +5,7 @@ const addImageLayer = (map, options, layer, nextLayer) => {
     layer.image,
     (err, image) => {
       if (err) { 
-        console.log(`Image ${layer.image} not found. Falling back to solid fill.`)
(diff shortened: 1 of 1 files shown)


app/javascript/components/map/mixins/mixin-add-layers.js
```diff
@@ -1,21 +1,29 @@
...
-
+const addImageLayer = (map, options, layer, nextLayer) => {
   map.loadImage(
     layer.image,
     (err, image) => {
-      if (err) { throw err }
-      
-      map.addImage(patternId, image)
+      if (err) { 
+        console.log(`Image ${layer.image} not found. Falling back to solid fill.`)
```
nym1k, line 8: Should this be console.error or console.warn?

nym1k (changes requested):

app/javascript/components/map/mixins/mixin-add-layers.js
WKocur: Hmmm potentially, yes... I think warn would be best as it fails gracefully

PR #174, opened 2021-06-08
Fix the next country page button

stanleypliu:
**Ticket**

https://unep-wcmc.codebasehq.com/projects/ocean-habitats/tickets/168

New scope in GeoEntity 
Use the same scope for populating the dropdown list of countries 
Should now wrap around (so Yemen will display a link to Albania)

### DIFF
--- a/app/controllers/countries_controller.rb
+++ b/app/controllers/countries_controller.rb
@@ -43,6 +43,10 @@ class CountriesController < ApplicationController
   private
 
   def next_country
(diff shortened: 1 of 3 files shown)


WKocur (commented):
Looks good - only small thing is that I would rename the allowed_countries scope to avoid naming conflict with ALLOWED_COUNTRIES and to reflect the fact that it's countries with stats. 

I've not seen the & operator before (had to look it up) - looks like a useful one!

stanleypliu:
Yep, `&` is intersection, `|` is union IIRC

PR #185, opened 2021-06-08
Fix importers

stanleypliu:
Changes:

- put habitat creation in a rake task
- invoke new habitat creation tasks in seeds and import:refresh task
- delete habitat and globalchangestat models in import:refresh

Works locally for me. Will deploy to staging if happy

**Notes**

Run `rake import:refresh` and it should be recreating the habitats

### DIFF
--- a/config/deploy/production.rb
+++ b/config/deploy/production.rb
@@ -1,5 +1,5 @@
 set :stage, :production
-set :branch, "supported-production"
+set :branch, "main"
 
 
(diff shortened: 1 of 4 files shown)


WKocur:
@stanleypliu Code looks good - just running to see if it works for me too. Do you think the issue with the stats on staging was due to the duplicate habitats?

WKocur:
Also, what's your plan for removing the other fixes? Will you do that after this is in develop?

stanleypliu:
There's no issue with the stats on staging, just on production. And yeah I need to do that after this is in develop - since there are bound to be conflicts!

PR #186, opened 2021-06-08
Release 1.1.0

stanleypliu:
Contains changes to the importers - will necessitate the `rake import:refresh` to be run on production after deploying.

### DIFF
--- a/app/models/geo_entity.rb
+++ b/app/models/geo_entity.rb
@@ -56,9 +56,9 @@ class GeoEntity < ApplicationRecord
 
     hash = {}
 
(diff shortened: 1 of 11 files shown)


WKocur:
Uniqs haven't been removed, just checking that's intentional? I can't see them doing any harm, but want to make sure you're happy with that.

Also I noticed this changed on main a couple of times: https://github.com/unepwcmc/ocean-plus-habitats/blob/c58e064788f280e6cf5098b7908f99b7d72f6eec/app/models/geo_entity.rb#L59-L63

Is it correct as it currently is on main?

PR #187, opened 2021-06-11
Fix/improve performance

stanleypliu:
**Ticket**

None

Simple fix to try and improve performance, particular when loading the homepage. This gets rid of an N+1 query when counting species by category by preloading the Habitat model, and it definitely reduces the number of SQL queries in the Rails log.

**Notes**

`bundle install` to install the Bullet gem.


WKocur:
@stanleypliu Not heard of this before - that's quite cool! I thought I'd give it a test to see the impact. Times are the time taken for localhost request to resolve for the homepage.

Tested and it takes the time for reload of the homepage down from 2-2.5s to 1s ish 👍 Noted that the page load is slower when bullet is enabled, so if refreshing a lot (e.g. for styling) people might want to disable bullet!

PR #195, opened 2021-08-12
Refactor/increase performance of global stats related methods

defaye:


app/models/geo_entity.rb
```diff
@@ -15,7 +15,7 @@ class GeoEntity < ApplicationRecord
 
   has_many :country_citations, foreign_key: 'country_id'
 
-  scope :countries, -> { where.not(iso3: nil || 'GBL') }
+  scope :countries, -> { where.not(iso3: nil).where.not(iso3: 'GBL') }
```
defaye, line 18: This is clearly a bug in the code as the developer must have intended to check for both nil or 'GBL' as writing nil OR 'GBL' makes no sense

PR #228, opened 2022-11-18
fix: fix source name on mangroves layer

WillKocur:
To confirm with Osgur whether this fix is better made in the web service rather than the code, as ideally source names won't change.

Putting this PR in just for visibility.


WillKocur:
Closing as Osgur is renaming on the arcgis side.

PR #231, opened 2022-11-28
Feat/pagination metadata

WillKocur:
Adds metadata to the index countries endpoint

### DIFF
--- a/app/controllers/api/v1/countries_controller.rb
+++ b/app/controllers/api/v1/countries_controller.rb
@@ -56,7 +56,7 @@ class Api::V1::CountriesController < Api::V1::BaseController
   end
 
   def total_count
-    @countries.count
+    @total_count ||= @countries.count
   end
 
   def page_count
--- a/app/controllers/api/v1/countries_controller.rb
+++ b/app/controllers/api/v1/countries_controller.rb
@@ -1,3 +1,5 @@
+PER_PAGE = 25.freeze
+
 class Api::V1::CountriesController < Api::V1::BaseController
   before_action :set_country, only: [:show]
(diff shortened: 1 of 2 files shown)


app/controllers/api/v1/countries_controller.rb
```diff
@@ -27,7 +29,39 @@ def set_country
...
+  def page
+    params[:page] || 1
+  end
+
+  def per_page
+    25
+  end
+
+  def total_count
+    @countries.count
```
sergiomarrocoli, line 61: could memoize or use ```.length``` here to avoid a database call

app/controllers/api/v1/countries_controller.rb
```diff
@@ -27,7 +29,39 @@ def set_country
...
+      total_count: total_count
+    }
+  end
+
+  def page
+    params[:page] || 1
+  end
+
+  def per_page
+    25
```
sergiomarrocoli, line 57: unless we're allowing a param here, should probably be a constant rather than a method

sergiomarrocoli (changes requested):
looks good. spec is broken but i think you already fixed them in the other PR. two minor comments, but otherwise happy to approve

app/controllers/api/v1/countries_controller.rb
sergiomarrocoli: could memoize with ```@total_count ||= @countries.count```

sergiomarrocoli (changes requested):
left a comment about memoization

WillKocur:
> left a comment about memoization

All done - thanks!

sergiomarrocoli (approved):

